PR #2, opened 2020-08-25
The brain seems fine, just pushing it up

vainaviv:


fazerlicourice7 (changes requested):
Can you delete the file that we don't need? They're both the same right?

PR #4, opened 2020-08-30
Brain can read floats from client. Take sensor_id as variable currently, not an input.

vainaviv:


Brain_I2C/Brain_I2C_Basic/Brain_I2C_Basic.ino
```diff
@@ -0,0 +1,26 @@
+#include <Wire.h>
```
fazerlicourice7, line 1: Delete duplicate file.

Brain_I2C/Brain_I2C_Basic.ino
```diff
@@ -0,0 +1,26 @@
...
+void setup() {
+  Wire.begin();        // join i2c bus (address optional for master)
+  Serial.begin(9600);  // start serial for output
+}
+
+void loop() {
+  int sensor_id = 0; //change to desired sensor_id
+  Wire.requestFrom(8, 4); // request 4 bytes from address 8
+  uint8_t index = 0;
+  union floatToBytes {
```
fazerlicourice7, line 12: move outside of loop. Define it globally.
Currently, it wastes a lot of memory recreating this object, when we can just modify the same one each time.

Brain_I2C/Brain_I2C_Basic.ino
```diff
@@ -0,0 +1,26 @@
+#include <Wire.h>
+
+void setup() {
+  Wire.begin();        // join i2c bus (address optional for master)
+  Serial.begin(9600);  // start serial for output
+}
+
+void loop() {
+  int sensor_id = 0; //change to desired sensor_id
+  Wire.requestFrom(8, 4); // request 4 bytes from address 8
```
fazerlicourice7, line 10: Instead of setting a specific sensor id in the loop, we want to ahve a global map of sensor id to I2c address and loop through each of those and do wire.request() and get data and stuff.

Brain_I2C/Brain_I2C_Basic.ino
```diff
@@ -0,0 +1,26 @@
...
+  uint8_t index = 0;
+  union floatToBytes {
+      char buffer[4];
+      float voltageReading;
+  } converter;
+  
+  while (Wire.available()){
+    if (sensor_id == 0) { //successive if statements for more IDs
+      converter.buffer[index] = Wire.read();
```
fazerlicourice7, line 19: So is the idea to have one converter for each sensor? OR use the same converter for multiple different sensors?
I think it makes sense to use the same one, and we can remove the if statement from here if we add a for loop at the top of the loop method that loops through all of our sensors.

Brain_I2C/Brain_I2C_Basic.ino
```diff
@@ -0,0 +1,26 @@
...
+  Serial.begin(9600);  // start serial for output
+}
+
+void loop() {
+  int sensor_id = 0; //change to desired sensor_id
+  Wire.requestFrom(8, 4); // request 4 bytes from address 8
+  uint8_t index = 0;
+  union floatToBytes {
+      char buffer[4];
+      float voltageReading;
```
fazerlicourice7, line 14: Once again, this is going to have to be more general. Some sensors will have several floats, most will have a different number of them. Use the max we expect to have (maybe 3, maybe more; not sure yet).

fazerlicourice7 (changes requested):
Let me know if you have any questions

Brain_I2C/Brain_I2C_Basic.ino
vainaviv: So should I have a globally defined dictionary that maps a sensor to a tuple (address of board, sensor ID on that board)? And then in the loop, just go through each of those sensors?

PR #5, opened 2020-09-04
Loops through all sensors and requests data. Gets back an array of fl…

vainaviv:
…oats.


fazerlicourice7 (approved):
Looks Great! :smile:

PR #9, opened 2020-09-18
packet's created, checksum implemented, sending data via RF

vainaviv:


Brain_I2C/Brain_I2C.ino
```diff
@@ -45,10 +50,50 @@ void loop() {
...
+  String packet = "{";
+  packet += (String)sensor.overall_id;
+  packet += ",";
+  char *data = 0;
+  int count = 0;
+  for (int i=0; i<6; i++) {
+    float reading = farrbconvert.sensorReadings[i];
+    if (reading > 0) {
+      packet += (String)reading;
+      data += (char)reading;
```
fazerlicourice7: so I'm not entirely sure this is right. 
what is a float parsed to a char?
31.2456 -> char =/= '31.24', It's not even equal to '3', I think You want to finish building packet as a string, and then look at every character within that.

Brain_I2C/Brain_I2C.ino
```diff
@@ -45,10 +50,50 @@ void loop() {
...
+    RFSerial.println(packet);
+    delay(sensor.clock_freq);
+  }
+}
+
+String make_packet(struct sensorInfo sensor) {
+  String packet = "{";
+  packet += (String)sensor.overall_id;
+  packet += ",";
+  char *data = 0;
```
fazerlicourice7: When initializing here, you need to specify size of the char array, I'm not exaclty sure what the max size would be 6* (4 digits + 2 decimal places) + 5 commas?

Brain_I2C/Brain_I2C.ino
```diff
@@ -45,10 +50,50 @@ void loop() {
...
-      Serial.print(farrbconvert.sensorReadings[i]);
-      Serial.print(" ");
+      float reading = farrbconvert.sensorReadings[i];
+      if (reading > 0) {
+      }
+    }
+    String packet = make_packet(sensor);
+    Serial.println(packet);
+    RFSerial.println(packet);
+    delay(sensor.clock_freq);
```
fazerlicourice7: With this, what will happen is that we'll be running at the speed of the slowest sensor since we'll be delaying by that amount each time. In this case, we'll check pressure and wait for 1 sec, and then check temp and wait for half a sec, so we still check pressure once for each time we check pressure.

What we want, is to check temp twice for every time we check pressure. So I think you should always delay by 100 or something reasonably small. Instead of clock_freq being a time to delay, make it a number denoting how often to check, for example 2 means every other, 5 means every 5th time. 
Add a loop counter that counts the number of iterations loop goes through, and if that counter % curr_freq == 0, then do the reading. else, do nothing. 
Every so often you can reset that counter to 0, probably choosing the largest number we have for curr_freq as the point at which to return to 0.

fazerlicourice7 (changes requested):

Brain_I2C/Brain_I2C.ino
vainaviv: Why do I have to specify it? You don't need to do that, it still compiles

fazerlicourice7: because you won't necessarily get all the memory you need. If this array needs to have 100 elements, but you define it like this, you will only have access to whatever memory is currently available. You'll need to specify the size of the array in order guarantee that you get enough space there.

Brain_I2C/Brain_I2C.ino
```diff
@@ -1,54 +1,146 @@
...
 
 void loop() {  
-  for (int j = 0; j < sizeof(all_ids)/sizeof(all_ids[0]); j++) {
+  for (int j = 0; j < sizeof(all_ids)/sizeof(sensorInfo); j++) {
+    if (sensor_checks[j][0] == sensor_checks[j][1]) {
+      sensor_checks[j][1] = 1;
+    } else {
+      sensor_checks[j][1] += 1;
+      delay(100);
+      continue;
```
fazerlicourice7: Say pressure should be read every 100 ms. , temp 200 ms, and GPS 3000 ms. Here all_ids[0] is the pressure, all_ids[1] is the temp, all_ids[2] is gps.
run through this code right here four times and tell me how often each of those sensors would read data.

You're very close, but not quite there. (the delay shouldn't be here, it should be outside of this for loop entirely)

fazerlicourice7 (changes requested):

Brain_I2C/Brain_I2C.ino
```diff
@@ -1,54 +1,146 @@
...
+  packet_content += ",";
+  for (int i=0; i<6; i++) {
+    float reading = farrbconvert.sensorReadings[i];
+    if (reading > 0) {
+      packet_content += (String)reading;
+      packet_content += ",";
+    } else {
+      packet_content.remove(packet_content.length()-1);
+      int count = packet_content.length();
+      char const *data = packet_content.c_str();
```
fazerlicourice7: Nice! This is exactly it.

Brain_I2C/brain_utils.h
```diff
@@ -0,0 +1,66 @@
...
+} farrbconvert;
+
+/*
+ * Data structure to store all information relevant to a specific sensor type.
+ */
+struct sensorInfo {
+  String sensor_name;
+  int board_address;
+  byte sensor_id;
+  int overall_id;
```
fazerlicourice7, line 19: What are the differences between these two?

fazerlicourice7: Oh nvm. I see.

fazerlicourice7 (approved):

PR #11, opened 2020-10-07
Finished code for receiving commands and calling appropriate methods.

vainaviv:


Brain_I2C/brain_utils.h
```diff
@@ -47,6 +60,82 @@ String make_packet(struct sensorInfo sensor) {
   return packet;
 }
 
+/*
+ * Decodes a packet sent from ground station in the following format:
+ * {<valve_ID>,<open(1) or close(0)|checksum}
+ * And calls the corresponding method in solenoids.h
+ */
+void decode_received_packet(String packet) {
```
fazerlicourice7: Instead of calling the solenoid function here, pass in a pointer to a valveInfo struct, and this function will populate that struct, and in Brain do the appropriate action

Brain_I2C/Brain_I2C.ino
```diff
@@ -47,7 +59,17 @@ void setup() {
   }
 }
 
-void loop() {  
+void loop() { 
+   String command = "";
+//  while (RFSerial.available()) {
+//    command += RFSerial.read();
+//  }
```
fazerlicourice7: I don't think this code will work. Can you add chars together to form a string?

fazerlicourice7 (changes requested):

Brain_I2C/brain_utils.h
vainaviv: I'm a little confused about what you mean by this. So do you mean in the loop in brain, decode what valve we are working with and pass that into this method? And what do you mean by "populate the struct"? Lastly, if Brain is doing the action, do you want this switch case statement to be in brain rather than brain_utils?

fazerlicourice7: The signature of this method should be `decode_received_packet(String packet, Valve_info *valve){ ... }` and instead of the switch case inside here, just populate the fields of the valve.
and then move the switch case to Brain

fazerlicourice7 (approved):
:tada:

fazerlicourice7:
Merged directly into master and E1

PR #14, opened 2020-10-12
Vamshi setup e1

fazerlicourice7:
E1 Code is ready for cold flow!!

### DIFF
--- a/Brain/Brain.ino
+++ b/Brain/Brain.ino
@@ -7,20 +7,13 @@
  */
 
 #include <Wire.h>
-//#include <SoftwareSerial.h>
 #include "brain_utils.h"
+#include <GPS.h>
+#include <Barometer.h>
+#include <ducer.h>
 
-//
-//SoftwareSerial RFSerial(2,3);
-//SoftwareSerial GPSSerial(7,8);
-
-//
 #define RFSerial Serial2
 #define GPSSerial Serial3
-
-// Initialize sensor libs
-//GPS gps(&GPSSerial);
-//Barometer baro(&Wire);
 
 
 // within loop state variables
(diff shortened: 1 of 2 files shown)


fazerlicourice7:
@andycate 
@taitmate38 
Also take a look

fazerlicourice7:
Depends on https://github.com/Space-Enterprise-at-Berkeley/Client-Side/pull/10

Brain/Brain.ino
```diff
@@ -50,22 +59,26 @@ valveInfo valve_ids[7] = {
...
+  Ducers::init(&Wire);
+  Barometer::init(&Wire);
+  GPS::init(&GPSSerial);
 }
 
-void loop() { 
+void loop() {
   String command = "";
 //  command = RFSerial.readString();
   command = "{26,1|456}";
```
ghost: I assume this won't be hard coded for flight, and this is just for testing?

Brain/Brain.ino
```diff
@@ -50,22 +59,26 @@ valveInfo valve_ids[7] = {
...
   int action = decode_received_packet(command, &valve);
   take_action(&valve, action);
```
ghost: I think we should make the decode_received_packet function & take action more general, to allow us to send other sorts of commands. We may only be opening valves right now, but it could be useful to be able to send other sorts of commands (such as for debugging, or to "initialize launch sequence") and easily add that in to the flight code. Maybe we can make a general decode received_packet, which could then pass a decoded packet to another function that will appropriately set the valve information, & likewise have a take_valve_action function or something like that, to allow similar type functions to be created for other messages & other purposes as they come up. Maybe this should be a larger conversation, but that's just what came to mind for me

Brain/Brain.ino
vainaviv: Yup this was just in there for testing purposes. I forgot to remove it before pushing. Vamshi can you please take it out.

Brain/Brain.ino
vainaviv: Yea that definitely makes sense! I think to make it a more general decode function, we would need to change the format of the packets received so different fields could represent different actions in some way. If I could have a little more clarity on what the additional actions we would want to take are, I can try to formulate a new packet construction and decoding scheme for that.

vainaviv (changes requested):
It looks great! Just remove the command string I used for debugging purposes.

Brain/Brain.ino
ghost: Yea, so something that I've seen work is that you have a document that defines what each packet will look like (like we have) and then you use that as the guide for writing functions. Such as, the document says that a solenoid command will have one field, whereas a different type of command may have 2 or 3 or however many fields. I know it may seem a little weird to be hard coding things like that, but it's fine since as long as it's based off a master reference document! For now, the other commands I can think of would be to arm the rocket and then initialize the launch sequence. We may want other commands to get telemetry that isn't regular, for debugging purposes (such as the battery voltage - not something that we want every second, but will ask for if we notice something strange happening.) Does that make sense? We can talk about it more at the meeting tomorrow too :)

fazerlicourice7: I'm going to go ahead and merge this in as is for now. You should move this conversation to a thread in the Avionics channel. Tag @evelynliu16  as well, since she's doing the encode/decode on ground station side.

vainaviv (approved):
Looks great!

PR #24, opened 2020-11-01
Vamshi/new therm

fazerlicourice7:
New thermocouple + fix parse packet errors

### DIFF
--- a/Brain/Brain.ino
+++ b/Brain/Brain.ino
@@ -35,7 +35,8 @@ sensorInfo sensors[numSensors] = {
   {"Temperature",                FLIGHT_BRAIN_ADDR, 0, 3}, //&(testTempRead)}, //&(Thermocouple::readTemperatureData)},
   {"All Pressure",               FLIGHT_BRAIN_ADDR, 1, 1},
   {"Battery Stats",              FLIGHT_BRAIN_ADDR, 2, 3},
-  {"Battery Stats",              FLIGHT_BRAIN_ADDR, 4, -1},
+  {"Solenoid Ack",               FLIGHT_BRAIN_ADDR, 4, -1},
+  {"Recovery Ack",               FLIGHT_BRAIN_ADDR, 5, -1},
 
   //  {"GPS",                        -1, -1, 7, 5, NULL}, //&(GPS::readPositionData)},
   //  {"GPS Aux",                    -1, -1, 8, 8, NULL}, //&(GPS::readAuxilliaryData)},
@@ -82,8 +83,7 @@ void loop() {
     }
     int action = decode_received_packet(String(command), &valve);
     take_action(&valve, action);
-    Solenoids::getAllStates(farrbconvert.sensorReadings);
-    make_packet(sensors[3]);
+    make_packet(valve.id); // this might need to be valve->id
   }
 
   /*
@@ -118,7 +118,7 @@ void loop() {
       // do nothing; these are on request packets
     }
 
-    String packet = make_packet(sensor);
+    String packet = make_packet(sensor.id);
     Serial.println(packet);
     RFSerial.println(packet);
     //bool did_write = write_to_SD(packet);
--- a/Brain/Brain.ino
+++ b/Brain/Brain.ino
@@ -26,21 +26,21 @@ uint8_t val_index = 0;
 String command = "";
 String file_name = "E1_" + String(month()) + "/" + String(day()) + "/" + String(year()) + "_" + String(hour()) + ":" + String(minute()) + ":" + String(second()) + "_log.txt";
 
-//Doesnt matter what goes here at initialization
-const char packet0[] PROGMEM = "Packet0";
-const char packet1[] PROGMEM = "Packet1";
-const char packet2[] PROGMEM = "Packet2";
-const char packet3[] PROGMEM = "Packet3";
-const char packet4[] PROGMEM = "Packet4";
-const char packet5[] PROGMEM = "Packet5";
-const char packet6[] PROGMEM = "Packet6";
-const char packet7[] PROGMEM = "Packet7";
-const char packet8[] PROGMEM = "Packet8";
-const char packet9[] PROGMEM = "Packet9";
+char packet0[] PROGMEM = "Packet0";
+char packet1[] PROGMEM = "Packet1";
+char packet2[] PROGMEM = "Packet2";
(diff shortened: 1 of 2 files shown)


Brain/brain_utils.h
```diff
@@ -87,22 +87,48 @@ String make_packet(struct sensorInfo sensor) {
...
+  Serial.println(checksum);
   
-  char const *data = packet.substring(1,data_end_index).c_str();
-  int count = data_end_index - 1; // sanity check; is this right? off by 1 error?
-  uint16_t check = Fletcher16((uint8_t *) data, count);
-  if (check == checksum) {
+  const int count = packet.substring(1, data_end_index).length(); // sanity check; is this right? off by 1 error?
+  Serial.println(count);
+  char data[count+1];// = packet.substring(1,data_end_index).c_str();
+  packet.substring(1, data_end_index).toCharArray(data, count + 1);
```
vainaviv, line 145: Just out of curiosity, was there something buggy with the previous way of getting the data?

Brain/brain_utils.h
```diff
@@ -87,22 +87,48 @@ String make_packet(struct sensorInfo sensor) {
...
+  
+  for (int i =0; i <= count; i++){
+    Serial.print(data[i]);
+  }
+  Serial.println();
+  uint16_t _check = Fletcher16((uint8_t *) data, count);
+  Serial.print("check 2: ");
+  Serial.println(_check);
+  Serial.println("got command");
```
vainaviv: Are these here for debugging, do we want to remove this?

Brain/Brain.ino
```diff
@@ -126,8 +126,8 @@ void loop() {
 void sensorReadFunc(int id) {
   switch (id) {
     case 0:
-      Thermocouple::setSensor(0);
-      Thermocouple::readTemperatureData(farrbconvert.sensorReadings);
+      //Thermocouple::setSensor(0);
+      Ducers::readTemperatureData(farrbconvert.sensorReadings);
```
vainaviv: I believe for case 1 and 2, you need to include the farrbconvert

vainaviv: Also I think it would make sense to move sensorReadFun to brain utils rather than cluttering the brain code

vainaviv (changes requested):

Brain/brain_utils.h
fazerlicourice7: it was doing some weird conversion. I think it had something to do with encoding, or maybe we were doing something incorrectly, lemme add it to trello as a task, someone should figure out why it wasn't working.

Brain/Brain.ino
fazerlicourice7: will move func to brain utils;
But what do you mean by the first statement?

Brain/brain_utils.h
vainaviv: Ok once I get the I2C working I can look into this also

Brain/Brain.ino
vainaviv: Look at the lines below. For case 0, once you got the data you did: 
farrbconvert.sensorReadings[1] = tempController::controlTemp(farrbconvert.sensorReadings[0]);
farrbconvert.sensorReadings[2] = -1;

Don't you need to do this for case 1 and case 2 or is it already doing that when you pass in farrbconvert.sensorReadings as a parameter

fazerlicourice7: It's done in the client side read functions.

fazerlicourice7:
need to test lates tomorrow before merging

Brain/Brain.ino
vainaviv: I see sounds good

vainaviv (approved):

Brain/Brain.ino
```diff
@@ -82,8 +88,12 @@ void loop() {
       i++;
```
vainaviv, line 92: Is the command array defined somewhere first? I don't see that happening anywhere

Brain/Brain.ino
```diff
@@ -82,8 +88,12 @@ void loop() {
...
     }
     int action = decode_received_packet(String(command), &valve);
-    take_action(&valve, action);
-    make_packet(valve.id); // this might need to be valve->id
+    if (action != -1) {
+      take_action(&valve, action);
+      String packet = make_packet(valve.id); // this might need to be valve->id
```
vainaviv: Just a note valve.id should be fine because id was just an int, not a pointer to a function.

Brain/brain_utils.h
```diff
@@ -132,7 +132,7 @@ int decode_received_packet(String packet, valveInfo *valve) {
   }
   int action = packet.substring(data_start_index + 1,data_end_index).toInt();
   
-  String checksumstr = packet.substring(data_end_index + 1, packet.length()-2);
+  String checksumstr = packet.substring(data_end_index + 1, packet.length()-1);
```
vainaviv, line 134: The reason it was -2 previously was because the packet was something like {1,2,3|5} so we didn't want the last '}' in the checksum. I believe this should stay -2 instead of -1 unless there was something changed about the packet format.

vainaviv (changes requested):
There are a couple comments I made. Is this finished or still being developed? I can help with testing and cleaning up the code if it's still in the develop phase.

Brain/Brain.ino
fazerlicourice7: line 26

Brain/Brain.ino
fazerlicourice7: yup. thanks

Brain/brain_utils.h
fazerlicourice7: That's what I thought too, but we were getting some errors this last weekend where the last value of the checksum was being chopped off. Will take note of this and watch out for it this saturday

fazerlicourice7: Worked perfectly with -1. 
makes sense since substring is exclusive of the end index
image

fazerlicourice7: Worked fine this weekend; this makes sense too since substring is exclusive of end index
image

fazerlicourice7:
Want to merge into E1 code w/out any inter board communication. Can talk more on Thursday meeting, but there's no way we're doing any interboard stuff for E1.

vainaviv (approved):

PR #26, opened 2020-11-25
constants seperate file, c_str, send message to GS when unable to write to SD

vainaviv:


Brain/Brain.ino
```diff
@@ -116,10 +98,10 @@ void loop() {
 }
 
 bool write_to_SD(String message){
-  // every reading that we get from sensors should be written to sd and saved.
```
fazerlicourice7, line 119: Instead of sending a string; can we add error packets to the [sheet](https://docs.google.com/spreadsheets/d/1xNmjrb5a28FfVg-MG9jUBRiPyfdkjo7HiqPWP41N_IQ/edit#gid=998950668) and send that instead.

Brain/brain_constants.h
```diff
@@ -0,0 +1,52 @@
...
+
+/*
+ * Data structure to store all information relevant to a specific valve.
+ */
+struct valveInfo {
+  String name;
+  int id;
+  int (*openValve)();
+  int (*closeValve)();
+};
```
fazerlicourice7, line 25: I think these should still be in utils, since they are structs, not really constants. This constants file should only contain things that might change (pin numbers, num sensors, array of sensors, etc).

Brain/brain_utils.h
```diff
@@ -6,6 +6,7 @@
  * Created by Vainavi Viswanath, Aug 21, 2020.
  */
 #include "solenoids.h"
+#include "brain_constants.h"
```
fazerlicourice7, line 9: Instead of including this here, can we have Brain.ino include both of them?

Brain/brain_utils.h
```diff
@@ -66,6 +38,11 @@ String make_packet(struct sensorInfo sensor) {
     packet_content.remove(packet_content.length()-1); 
     int count = packet_content.length();
     char const *data = packet_content.c_str();
+    // c_str converts the String to it's underlying c-string type array and returns a
+    // const char* to avoid tampering of the data. The toCharArray makes a copy of the 
+    // string and you have to specify a length of the elements copied. Previously may not
+    // have been working due to length variations or possibly incorrect copying, not enough
+    // space to copy.
     uint16_t checksum = Fletcher16((uint8_t *) data, count);
     packet_content += "|";
```
fazerlicourice7, line 47: So, re c_str(), the issue we had was w/ `decode_received_packet`, not `make_packet`
If you take a look at the E1 branch, this func is very very similar, but the decode func is heavily modified. 
And we're decoding the checksum with c-str / toCharArray, so it's always constant size, so I'm not sure this is the issue.

fazerlicourice7 (changes requested):

fazerlicourice7:
Before we make a lot of functional changes here, I think we need to update master w/ the changes made to E1. Doing that first will show us what really needs to be done here.
Would be great if you could do that. But really, I think this is not efficient. I don't think my original idea of diff branches for diff versions works that well. Could you do some reading on how better to organize the code so that we can easily adapt it to different hardware (the first thing is multiple config files for E1-horfire, E1-coldflow, E1-flight, LAD4, but also something that will allow us to deal w/ different versions of hardware, eg v1.0 pcb, v2.0 pcb, etc.

One thing we're planning on doing is merging client and Brain into on repo (over winter break), and then a lot of this will be easier to manage.

But we still need a way to organize this stuff. Is it just multiple different folders? ie. a folder for E1 w/ 3 diff config files, A folder for LAD4 w/ x diff config files? OR what? And what about when we add a new board + code, is the workflow to copy an old folder and modify it, or should we have some sort of template like we do now? and if we do have a template, how to ensure that it gets all the updates that the flight specific code gets (issue we have now). additionally, when we have multiple diff boards + code, how to ensure updates to one get applied to others (if applicable).
These are all things we need to figure out. Would be great if you could think about them/ do some reading. Don't need to figure it all out.

fazerlicourice7:
We can merge this branch w/ constants file and sd fail rf write stuff.
rename the constants file to config or something.

fazerlicourice7:
Idk if you have time to talk about this stuff over the next two weeks, I definitely don't think I was able to convey what I wanted to in that paragraph. If not, we can talk after finals; this will be a great winter break project for us.

Brain/brain_constants.h
vainaviv: Yea but the tricky thing is then I can't import brain utils in both brain constants and brain.ino because it says the variables were initialized twice. I think can use the extern keyword though. Will check this out

vainaviv:
Yea I might have time over the next two weeks to talk about this, but I'm not sure if I will have time to implement anything until after finals. I'll let you know a time.

PR #27, opened 2020-12-02
Fix writing to SD

fazerlicourice7:
Very fast ~100us per write
Queue has been exhaustively tested. No memory leaks!!

### DIFF
--- a/Brain/Brain.ino
+++ b/Brain/Brain.ino
@@ -13,9 +13,6 @@
 
 #define RFSerial Serial6
 #define GPSSerial Serial8
-#include <SD.h>
-#include <SPI.h>
-#include <TimeLib.h>
 
 #define FLIGHT_BRAIN_ADDR 0x00
 
@@ -57,12 +54,15 @@ int sensor_checks[numSensors][2];
 valveInfo valve;
 sensorInfo sensor = {"temp", 23, 23, 23};
 
+std::string str_file_name = "E1_speed_test_results.txt";
+const char * file_name = str_file_name.c_str();
+
 long startTime;
 String packet;
 
 void setup() {
   Wire.begin();
-  Serial.begin(9600);
+  Serial.begin(57600);
   RFSerial.begin(57600);
   GPSSerial.begin(4608000);
 
@@ -72,6 +72,15 @@ void setup() {
     sensor_checks[i][0] = sensors[i].clock_freq;
     sensor_checks[i][1] = 1;
   }
+
+  int res = sd.begin(SdioConfig(FIFO_SDIO));
+  if (!res) {
+      // make packet with this status. print, and send over rf
+      // do additional error handling; what if no sd?
+  }
+  file.open(file_name, O_RDWR | O_CREAT);
+  file.close();
+  sdBuffer = newQueue();
 
   Recovery::init();
   Solenoids::init();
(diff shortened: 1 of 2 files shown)


Brain/Brain.ino
```diff
@@ -119,49 +131,32 @@ void loop() {
...
-
-  if (bfr_idx == 10) {
-    File myFile = SD.open(file_name.c_str(), FILE_WRITE);
-
-    if (myFile) {                                                     //If the file opened
-      for (int i = 0; i <= bfr_idx; i++) {
-        strcpy_P(buffer, (char *)pgm_read_word(&(packet_table[i])));  // Necessary casts and dereferencing, just copy.
-        myFile.println(buffer);
+    sdBuffer->enqueue(message);
+    if(sdBuffer->length >= 40) {
```
vainaviv, line 147: Is there a reason why >= 40. Is this the similar logic to before where you only write after 10 packets have been written onto?

Brain/Brain.ino
```diff
@@ -119,49 +131,32 @@ void loop() {
...
+      if(file.open(file_name, O_RDWR | O_APPEND)) {
+        int initialLength = sdBuffer->length;
+        for(int i = 0; i < initialLength; i++) {
+          char *msg = sdBuffer->dequeue();
+          file.write(msg);
+          free(msg);
```
vainaviv, line 153: In utils, you commented that whenever dequeueing, be careful and remember to clear string after calling dequeue. I assume this is what this line is doing?

Brain/brain_utils.h
```diff
@@ -12,10 +12,100 @@
...
+      tmp->next = nullptr;
+      free(tmp);
+
+      return msg;
+    }
+    return nullptr;
+  }
+};
+
+struct Queue * newQueue() {
```
vainaviv: I'm a little confused as to what the purpose of newQueue is. It looks like you're using it as a constructor, is that correct?

vainaviv (changes requested):
Hey, I looked through it and it looks pretty good! I just had a couple questions :)

Brain/Brain.ino
fazerlicourice7: Correct. I can't free the mem within the dequeue method because I want to access that string wherever I call dequeue, so right now I'm just counting on the user to properly free the mem, hence the comment reminding them to do so.

Brain/brain_utils.h
fazerlicourice7: exactly. Just a small method that allocates mem and initializes var

Brain/Brain.ino
fazerlicourice7: yeah. if we have 40 packets in the queue (in memory), we open the sd and write all 40 at once. This is because opening and closing the sd is pretty slow. 40 is kind of arbitrary; we're optimizing keeping the queue as small as possible (so that we don't lose data in case of crash) and super fast writes which means very infrequent sd open/close, which means very large queue; as you can see they're inversely related.

fazerlicourice7:
Document with some interesting phenomenon encountered here: https://docs.google.com/document/d/1Iy9Oyh2XKZbtX3eHR8_S-tz7GyAjmGOp3HR4DR_dFMg/edit

vainaviv (approved):
Looks great! 😍

PR #28, opened 2020-12-20
fixed issue with c_str not working

vainaviv:


Brain/Brain.ino
```diff
@@ -84,16 +84,19 @@ void setup() {
...
       Serial.print(command[i]);
       i++;
     }
     Serial.println();
-    Serial.println(String(command));
-    int action = decode_received_packet(String(command), &valve);
+    //Serial.println(String(command));
+    //int action = decode_received_packet(String(command), &valve);
+    Serial.println("{20,12|12345}");
+    int action = decode_received_packet("{20,12|12345}", &valve);
```
fazerlicourice7: Can you revert these?
And the checksum should be 4 digits. Can you verify this works w/ a 4 digit checksum? try something hex too

vainaviv: Reverted back and also tested that it does work with 4 digit hex numbers

fazerlicourice7 (approved):
Fix the merge conflicts, and your good to go :ok-hand:

PR #29, opened 2020-12-23
added error packet for when unable to write to SD

vainaviv:


vainaviv:
I updated the E1 Design spreadsheet and included the codes for error packets and assigned SD write error to 101. All errors will be in the 100s

Brain/Brain.ino
```diff
@@ -58,6 +58,7 @@ const char * file_name = str_file_name.c_str();
 
 long startTime;
 String packet;
+String error_packet;
```
fazerlicourice7: don't think we need a separate variable for this; just use packet. It's a transient variable anyway, it changes on every iteration

fazerlicourice7 (commented):
just one tiny change. And hold off on merging for a sec. Let's see what the resolution of discussion in slack is.

Brain/Brain.ino
```diff
@@ -128,12 +132,9 @@ void loop() {
     board_address = sensor.board_address;
     sensor_id = sensor.id;
     sensorReadFunc(sensor.id);
-    packet = make_packet(sensor.id);
+    packet = make_packet(sensor.id, false);
     Serial.println(packet);
     RFSerial.println(packet);
-    if (!write_to_SD(packet.c_str())){
```
fazerlicourice7, line 134: Still need to actually writeToSD, just don't do anything w/ the results

fazerlicourice7:
and fix merge conflicts

Brain/Brain.ino
vainaviv: Oops added that back in

PR #32, opened 2020-12-25
Vamshi/adc libs refactor

fazerlicourice7:
I've started the fw refactor we talked about. Sandboxing w/ only ADC related libs (analog thermocouple and pressure transducers).
Here I've finished the analog thermocouple refactor, and am putting it out there for prelim review. 

This hasn't been tested yet. This code compiles, but won't work until I finish the ducers lib refactor.

### DIFF
--- a/E1_coldflow/brain_utils.h
+++ b/E1_coldflow/brain_utils.h
@@ -31,20 +31,12 @@ int ADSAddrs[numADCSensors] = {0b1001010, 0b1001000};
 int adcDataReadyPins[numADCSensors] = {29, 28};
 ADS1219 ** ads;
 
-// sparse matrix representation of mapping
-int adc1ThermMap[] = {0, 0, 0, 0};
-int adc2ThermMap[] = {0, 0, 1, 0};
-int * adcThermMap[] = {adc1ThermMap, adc2ThermMap};
-
-std::vector<int> *analogThermADCMap;
+const int numAnalogThermocouples = 1;
+int thermAdcIndices[numAnalogThermocouples] = {1};
+int thermAdcChannels[numAnalogThermocouples] = {2};
 
 void initConfig() {
-  analogThermADCMap = (std::vector<int> *)malloc(numADCSensors * sizeof(std::vector<int>));
-  for (int i = 0; i < numADCSensors; i++) {
-    // initialize vector w/ iterator to beginning and end of array
-    analogThermADCMap[i] = std::vector<int>(adcThermMap[i], adcThermMap[i] + sizeof(adcThermMap[i]) / sizeof(int) );
-    analogThermADCMap[i].shrink_to_fit();
-  }
+
 }
 
 // SD_FAT_TYPE = 0 for SdFat/File as defined in SdFatConfig.h,
--- a/E1_coldflow/brain_utils.h
+++ b/E1_coldflow/brain_utils.h
@@ -16,19 +16,36 @@
 #include <string>
 #include <SdFat.h>
 #include <TimeLib.h>
+#include <vector>
 
 String make_packet(struct sensorInfo sensor);
 uint16_t Fletcher16(uint8_t *data, int count);
 void chooseValveById(int id, struct valveInfo *valve);
 
 const int numCryoTherms = 2;
-int cryoThermAddrs[numCryoTherms] = {0x60, 0x67}; //the second one is 6A or 6B, not sure which for Addr pin set to 1/2
+int cryoThermAddrs[numCryoTherms] = {0x60, 0x67};
 _themotype cryoTypes[numCryoTherms] = {MCP9600_TYPE_J, MCP9600_TYPE_T};
 
 const int numADCSensors = 2;
 int ADSAddrs[numADCSensors] = {0b1001010, 0b1001000};
 int adcDataReadyPins[numADCSensors] = {29, 28};
 ADS1219 ** ads;
+
+// sparse matrix representation of mapping
+int adc1ThermMap[] = {0, 0, 0, 0};
+int adc2ThermMap[] = {0, 0, 1, 0};
+int * adcThermMap[] = {adc1ThermMap, adc2ThermMap};
+
+std::vector<int> *analogThermADCMap;
+
+void initConfig() {
+  analogThermADCMap = (std::vector<int> *)malloc(numADCSensors * sizeof(std::vector<int>));
+  for (int i = 0; i < numADCSensors; i++) {
+    // initialize vector w/ iterator to beginning and end of array
+    analogThermADCMap[i] = std::vector<int>(adcThermMap[i], adcThermMap[i] + sizeof(adcThermMap[i]) / sizeof(int) );
+    analogThermADCMap[i].shrink_to_fit();
+  }
+}
 
 // SD_FAT_TYPE = 0 for SdFat/File as defined in SdFatConfig.h,
 // 1 for FAT16/FAT32, 2 for exFAT, 3 for FAT16/FAT32 and exFAT.
--- a/E1_coldflow/brain_utils.h
+++ b/E1_coldflow/brain_utils.h
@@ -176,8 +176,6 @@ void sensorReadFunc(int id) {
       farrbconvert.sensorReadings[2] = -1;
       break;
     case 1:
-      Serial.println("reading pressures");
-      Serial.flush();
       Ducers::readAllPressures(farrbconvert.sensorReadings);
       break;
     case 2:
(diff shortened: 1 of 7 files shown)


E1_coldflow/brain_utils.h
```diff
@@ -16,15 +16,37 @@
...
+
+void initConfig() {
+  analogThermADCMap = (std::vector<int> *)malloc(numADCSensors * sizeof(std::vector<int>));
+  for (int i = 0; i < numADCSensors; i++) {
+    // initialize vector w/ iterator to beginning and end of array
+    analogThermADCMap[i] = std::vector<int>(adcThermMap[i], adcThermMap[i] + sizeof(adcThermMap[i]) / sizeof(int) );
+    analogThermADCMap[i].shrink_to_fit();
+  }
+}
+
```
fazerlicourice7: This will be the general layout of the config files. Open to ideas to make it cleaner, but given that everything here is by definition hardcoded, not sure what we can do here.

libraries/thermocouple/Thermocouple.h
```diff
@@ -97,71 +100,95 @@ namespace Thermocouple {
...
+
+    long rawRead;
+    float voltageRead;
+    float tempRead;
+
+    void init (int numSensors, vector<int> * analogInMap, ADS1219 ** adcs) {
+      _numSensors = numSensors;
+      _analogInNumMap = analogInMap; // since this variable is coming from the config, I don't think I need to copy it, since it's not going to change/ be deallocated
+      _adcs = _adcs;
+    }
```
fazerlicourice7: library is initialized w/ all maps: i2c addresses , pin defs (no pins here), any other data.

libraries/thermocouple/Thermocouple.h
```diff
@@ -97,71 +100,95 @@ namespace Thermocouple {
...
+
+    // float *data is only of size 7 rn, ensure that we only expect <= 7 readings.
+    void readTemperatureData(float *data) {
+      int index = 0;
+      for (int i = 0; i < _numSensors; i++) {
+        int ainCounter = 0;
+        for (std::vector<int>::iterator it = _analogInNumMap[i].begin();
+             it != _analogInNumMap[i].end(); ++it ) {
+          if(*it != 0) {
```
fazerlicourice7: generalized data reads in the library. 

no reference to specific locations / functions of thermocouple here. That should all be maintained w/in config

fazerlicourice7 (commented):
One implementation detail here:
All adc objects will go to all libs that need *any* adc objects, and the sparse map of which analog inputs are used will define which ads objects end up getting used. 

This seemed cleaner to me than multiple different arrays w/ pointers to the same ADC objects
ie. something like
```
for (int i = 0; i < numADCSensors; i++) {
    ads[i] = new ADS1219(adcDataReadyPins[i], ADSAddrs[i], &Wire);
  }

thermAds[] = {&ads[0]};
ducerADC[] = {&ads[0], &ads[1]};
```

E1_coldflow/brain_utils.h
```diff
@@ -16,15 +16,37 @@
...
+#include <vector>
 
 String make_packet(struct sensorInfo sensor);
 uint16_t Fletcher16(uint8_t *data, int count);
 void chooseValveById(int id, struct valveInfo *valve);
 
 const int numCryoTherms = 2;
-int cryoThermAddrs[numCryoTherms] = {0x60, 0x67}; //the second one is 6A or 6B, not sure which for Addr pin set to 1/2
+int cryoThermAddrs[numCryoTherms] = {0x60, 0x67};
 _themotype cryoTypes[numCryoTherms] = {MCP9600_TYPE_J, MCP9600_TYPE_T};
```
taitmate38, line 26: missing an r - typo? should be `_thermotype`

fazerlicourice7: unfortunately not lol. the author of that lib has the typo

fazerlicourice7: but any comments on the general structure implementation?

vainaviv:
Can we move the write_to_SD to the brain utils.

vainaviv:
Weren't we going to split up brain_utils. Right now it seems to have both the constants and the stuff that should be in common_firmware and the generalized commands library. Do we want to have all constants in a completely separate file, not even in brain_utils. It just seems a bit long and for me personally, hard to understand.

vainaviv (changes requested):
See my other two comments.

fazerlicourice7:
Should've made this more clear. This is not ready to actually be merged in. I'm only getting started on the refactor. 
I just wanted to open this up to review before I refactored everything. Specifically see the top of `brain_utils.h` that I commented on that outlines what the information in the config file will look like.
And look at the `Thermocouples::Analog` & `Thermocouple::Cryo` library to see what the libraries will look like. 
I'm looking for feedback / comments on those

vainaviv (approved):

E1_coldflow/brain_utils.h
taitmate38: What exactly is the advantage of using the vectors here? At the end of the day they're just used to iterate through each ADC map once inside the library, right?

libraries/thermocouple/Thermocouple.h
taitmate38: I guess one concern I have is about the latency that this will add to each sensor reading, since we'll have to iterate through all ADC channels every time. Especially if we add more ADCs, this would add unnecessary delay. Since upon initialization we'll know how many of each sensor we have, at the beginning we could iterate through the sparse ADC mapping once and save the ADC # & ch # for each sensor, then when reading data just use those, and save doing the unnecessary iteration on every read

taitmate38: Or you could create the 2 arrays with the ADC # & ch # for each sensor in the config, and just pass those into the library, so that when you need to read in the library you can just iterate over the # of sensors & Ch ___ of ADC ___, and only have to do that the same # of times as sensors you have.

E1_coldflow/brain_utils.h
taitmate38: I think it could be simpler to pass an array to each type of sensor, cause then you could avoid the complexity & memory overhead of vectors. Instead of passing an array with pointers to each library, you could instead just have one global adc array, and pass to each library an array containing indices for that global array, and the channel number for each sensor. Something like this:


```
for (int i = 0; i < numADCSensors; i++) {
    ads[i] = new ADS1219(adcDataReadyPins[i], ADSAddrs[i], &Wire);
  }

int numTherm = 2;
int thermAdcs[] = {0, 1};
int thermAdcCh[] = {0, 3};

int numPT = 4;
int ptAdcs[] = {0, 0, 1, 1};
int ptAdcChs[] = {1, 2, 0, 1};
```
-----

Then in the library itself you could just iterate over the # of sensors:

```
init (int numSensors, int *adcs, int * chs) {
      _numsensors = numSensors;
      _adcs = ADCs;
      _chs = chs;
      
}

read_data {
      for (int i = 0; i < _numSensors; i++) {
            rawRead = _adcs[i]->readData(chs[i]);
            /*Sensor Specific stuff*/
      }
}
```

Any thoughts/hesitations with an approach like this?

fazerlicourice7: nice! I was originally trying something like this, but I kept trying to keep the map 2d and was having issues. Didn't think about flattening it like this. Will implement and rerequest review

E1_coldflow/brain_utils.h
```diff
@@ -22,9 +22,26 @@ uint16_t Fletcher16(uint8_t *data, int count);
...
+ADS1219 ** ads;
+
+const int numAnalogThermocouples = 1;
+int thermAdcIndices[numAnalogThermocouples] = {1};
+int thermAdcChannels[numAnalogThermocouples] = {2};
+
+const int numPressureTransducers = 5;
+int ptAdcIndices[numPressureTransducers] = {0, 0, 0, 0, 1};
+int ptAdcChannels[numPressureTransducers] = {0, 1, 2, 3, 0};
+
```
fazerlicourice7, line 40: Flattened ADC->channel maps

libraries/thermocouple/Thermocouple.h
```diff
@@ -97,71 +100,88 @@ namespace Thermocouple {
...
+    // float *data is only of size 7 rn, ensure that we only expect <= 7 readings.
+    void readTemperatureData(float *data) {
+      int i = 0;
+      while (i < _numSensors) {
+        rawRead = _adcs[_adcIndices[i]]->readData(_adcChannels[i]);
+        voltageRead = (float) rawRead * (5.0 / pow(2,23));
+        data[i] = ((voltageRead - voltageOffset) * tempOverVoltageScale) + tempOffset;
+        i++;
+      }
+      data[i] = -1;
```
fazerlicourice7, line 136: read function is linear in number of sensors of this type instead of linear in total number of Adcs.

fazerlicourice7:
Can @andycate  or @taitmate38  test this code on the PCB that either of you have? if it works, I think this is ready to merge in.

libraries/thermocouple/Thermocouple.h
taitmate38: Changes look great! All around looks pretty clean & understandable, both in config file & in library too. I'll test this build out on the Flight Computer that I have

fazerlicourice7:
Ben and I tested this this morning! Everything works great!

PR #33, opened 2020-12-30
Generalizing commands

vainaviv:
I worked on taking commands to request specific sensor/battery data. It can now receive those commands and properly provide that data. I definitely have not finished generalizing all the commands. I am planning on making two groups: request_data and perform_action. The "sensors" are right now in place for the request_data group and the "valves" are in place for the perform_action group. I just wanted you to take a quick look before I proceeded farther.


fazerlicourice7:
Can you rename the file `E1_Coldflow/Brain/Brain.ino` to `E1_Colflow/E1_Coldflow.ino`, the way it's done here, It's not obvious what the diffs are, It'll be much easier to see if I have the diff view that git provides
delete whatever is inside the `E1_Coldflow/Brain` dir please.

E1_coldflow/brain_utils.h
```diff
@@ -151,31 +182,49 @@ valveInfo valves[numValves] = {
...
+ * action in solenoids.h
+ */
+void take_action(valveInfo *valve, sensorInfo *sensor, int action) {
+  if (action == 1) {
+    valve->openValve();
+  } else if (action == 0) {
+    valve->closeValve();
+  } else if (action == 2) {
+    //call the reading function
+    sensor->readData(farrbconvert.sensorReadings);
```
fazerlicourice7, line 224: So the reason we had the switch case was that this never worked in the past. This needs to be tested more in an isolated way w/ minimal sensors.
And this looks good for requesting specific data, but I'm not sure I see where you handle the regular data. There will still be certain data that we want to just stream w/out receiving a command for.

E1_coldflow/brain_utils.h
```diff
@@ -216,21 +265,25 @@ String make_packet(int id, bool error) {
 
 /*
  * Decodes a packet sent from ground station in the following format:
- * {<valve_ID>,<open(1) or close(0)|checksum}
+ * {<ID>,<command (optional)>|checksum}
  * Populated the fields of the valve and returns the action to be taken
  */
-int decode_received_packet(String packet, valveInfo *valve) {
+int decode_received_packet(String packet, valveInfo *valve, sensorInfo *sensor) {
   Serial.println(packet);
```
fazerlicourice7, line 272: And does this move to the commands library? OR are you planning on keeping it on Common Fw?

E1_coldflow/brain_utils.h
```diff
@@ -151,31 +182,49 @@ valveInfo valves[numValves] = {
...
+/*
+ * Calls the corresponding method for this valve with the appropriate
+ * action in solenoids.h
+ */
+void take_action(valveInfo *valve, sensorInfo *sensor, int action) {
+  if (action == 1) {
+    valve->openValve();
+  } else if (action == 0) {
+    valve->closeValve();
+  } else if (action == 2) {
```
fazerlicourice7, line 222: I think we should use enums here instead of just hardcoding the value.

fazerlicourice7 (commented):
Made some comments.
So the way I'm understanding it, a command can either influence a valve, or a sensor right? And if it's a valve, its a `do something` command, and if it's a sensor, it's a `get Data` command? 
Currently the decode func takes a sensor struct and a valve struct and it looks like it only uses one or the other based on what type of command it is. I think we should try to generalize that if possible so it only takes one object, and can populate it appropriately based on whether its a sensor or a valve (or something else, I think the concept of a valve needs to be generalized). Additionallly, we could do something like an abstract struct that both valve and sensors extend so that either a sensor or valve can be represented by the same base struct

PR #34, opened 2020-12-31
Kalman filter Based Apogee Detection

fazerlicourice7:
The output of the Apogee detection algorithm in C++:
kalman_cpp
The output of the apogee detection algorithm built in python: As you can see on the graph, the choice of apogee is very reasonable.
kalman_notebook
The `42657` in the jupyter notebook is the time in ms, which is equal to the `42.6575s` found by the C++ code.

The Python code (exact same algorithm) has also been tested on LAD4 data and works as expected:
kalman_LAD4_data

This PR is based on #32 , so only merge this after that is merge in.


fazerlicourice7:
Some of the changes you see here are from the other branch, feel free to ignore those / review them in the other PR.

E1_coldflow/brain_utils.h
```diff
@@ -182,7 +182,7 @@ void sensorReadFunc(int id) {
       batteryMonitor::readAllBatteryStats(farrbconvert.sensorReadings);
       break;
     case 4:
-      Thermocouple::Cryo::readCryoTemp(farrbconvert.sensorReadings);
+      Thermocouple::Cryo::readCryoTemps(farrbconvert.sensorReadings);
```
vainaviv: The switch case is going to change soon here, just pointing it out to keep in mind

vainaviv (approved):

PR #35, opened 2021-01-05
Move to PlatformIO build system

andycate:
All the dependencies are working now, but the kalman filter seems to have bugs which need to be fixed.


taitmate38:
We should move some of the other .cpp files that are currently in the src direction into `lib` and make them libraries, since they are libraries. imo src should only have what is/used to be in `brain_utils` and `brain.ino`. It should be as simple as moving things around to have the right file format, as described in the `lib` readme, right?
|--lib
|  |
|  |--Bar
|  |  |--docs
|  |  |--examples
|  |  |--src
|  |     |- Bar.c
|  |     |- Bar.h

That way we can maintain the directory structure we were starting where we have separate directories for different versions of the flight code, between water flow, cold flow, hot fire, and flight

andycate:
Ok sure that makes sense. I can do that.

fazerlicourice7:
bruh don't add all of eigen to our repo. That thing is gigantic. And this is terrible practice.
Add that back to the gitignore
Install instructions are in the readme

fazerlicourice7:
and elaborate on kalman filter bugs. Is it still eigen related?

fazerlicourice7:
nvm. I found them.

fazerlicourice7:
So does it always compile everything? Where are the configs to tell it which main file it should use, so we can select `E1_coldflow` vs `E1_hotfire` at compile time

fazerlicourice7:
And can you add a quickstart to the top level read me

fazerlicourice7:
Just renaming the vars in kalman fixed the compile error. Is platformio enforcing some conventions? Same thing works fine w/g++

Will push later tonight

fazerlicourice7:
For Eigen: renaming all the files and changing the structure around is a drop. Figure out the library.json file and set it up so that we can just drop in an eigen folder and it'll work. having to rename the files is not the way we want to set it up. We have this one opportunity to do it right. If we don't do it right now, we never will.

fazerlicourice7:
And what's the status on this? I've been waiting on this to finish my refactors. Should I just continue those off master? Or are you close to done?

andycate:
I kinda disagree on removing the Eigen library. There are a few benefits to not doing that:
- Having super fast setup times
- Easy for new members
- Reproducible builds (we don't have to worry about people installing different versions)
Do you still want to do it? If so I can write up the instructions like you said

fazerlicourice7:
First, the structure w/in the eigen dir is not one that I've seen in an eigen pakcage that I downloaded. It looks like you moved stuff around and renamed stuff. At the very least, it should be plug and play to update that. I want to be able to drop in a new eigen package as downloaded from their site and have it work. Let's fix that first. 

Re having egein in the repo, I'm going to do some reading about it. we might be able to just throw in some sort of copyright stuff and it'll be fine. But I'm still kinda sussed about that

andycate:
I didnt change any of the file names or structures. What I did was copy paste the folder named `Eigen` from the downloaded release. 
image

andycate:
Also, the copyright stuff isn't a problem as long as we don't try to sell this code I'm pretty sure

fazerlicourice7:
Alright. I'm on board. 
Just copy all the files called `COPYING.*` into the Eigen folder. I think that's all we need to do. 

Fix merge conflicts and we're good to go!

fazerlicourice7 (approved):

andycate:
Sounds good!

PR #37, opened 2021-01-14
let cryo thermocouple work even if no sensor is detected

fazerlicourice7:
@benjamint8  try this.


taitmate38:
Didn't work - will still hang if trying to start without a thermocouple amplifier. This time, it got stuck during the library initialization in `setup`

fazerlicourice7:
`Serial.println("Error initializing cryo board at Addr 0x" + String(addrs[i], HEX));
          Serial.println("Continuing anyway. Will get garbage data.");`
Did you see these?

fazerlicourice7:
@benjamint8 bump in case you forgot (not urgent)

taitmate38:
this is not good to go, right? Still chasing down the memory leak/error

fazerlicourice7:
Decided we don't need this since we introduced the config -> hardware versioning. Created an `E1_waterflow` version instead that didn't use the cryo thermocouples.

PR #38, opened 2021-01-17
Vamshi/brain refactor

fazerlicourice7:
Brain refactor;
constants -> `config.h`
common code -> `common_fw`
specific code -> `___.cpp`

issue with pio though:
`common_fw.h` should go in `include/` and `common_fw.cpp` should go in `src`, but it fails to find all the relevant files; and stuff in common_fw doesn't exist. 
But if I put both in a lib folder, it works fine.

Doesn't work:
Screenshot from 2021-01-17 10-41-51

Works:
Screenshot from 2021-01-17 10-42-30

I don't care all that much about this diff, so I'm fine to merge in as a lib, but if you care, feel free to figure that out. 

This should be tested before merging.


andycate:
If you want to import files from the `lib` directory, you use `<>`. If you want to import files from `include` then you use `""`.

fazerlicourice7:
@andycate that doesn't work. (I'm assuming the .cpp file corresponding to the .h in include goes in src right?) It doesn't work.

fazerlicourice7:
This has gotten pretty fucking fat... let's hope there aren't any errors, or it will be a lot of fun to test... I'd recommend doing some initial checks before the weekend if possible for ^^ that reason.

src/E1_coldflow/E1_coldflow.cpp
```diff
@@ -12,7 +12,13 @@
...
 #include <batteryMonitor.h>
 
-#define RFSerial Serial6
+#define SERIAL_INPUT 1
+
+#if SERIAL_INPUT
+  #define RFSerial Serial
+#else
+  #define RFSerial Serial6
+#endif
```
fazerlicourice7, line 21: Was radio not working today?

taitmate38 (approved):
tested & doesn't blow up everything.

PR #39, opened 2021-01-31
new packet format. encode & decode

fazerlicourice7:


fazerlicourice7:
Needs to be tested. It compiles. 
Need corresponding code on GS side.

lib/common_fw/common_fw.cpp
```diff
@@ -5,10 +5,15 @@
  */
 #include "common_fw.h"
 
+#define PACKET_START 0x2A
+#define PACKET_END 0xC5
```
andycate: Are these choices arbitrary?

lib/common_fw/common_fw.cpp
```diff
@@ -36,36 +41,34 @@ bool write_to_SD(std::string message, const char * file_name) {
  * Constructs packet in the following format:
  * {<sensor_ID>,<data1>,<data2>, ...,<dataN>|checksum}
  */
-String make_packet(int id, bool error) {
-  String packet_content = (String)id;
-  packet_content += ",";
+uint8_t make_packet(uint8_t id, bool error) {
+  int i = 0;
+  packet[i++] = (PACKET_START >> 8) & 0xFF;
+  packet[i++] = PACKET_START & 0xFF;
```
andycate, line 52: Does this send the PACKET_START byte forwards and backwards?

lib/common_fw/common_fw.cpp
```diff
@@ -36,36 +41,34 @@ bool write_to_SD(std::string message, const char * file_name) {
...
+
+  packet[i++] = id;
   if (!error) {
-    for (int i=0; i<7; i++) {
-      float reading = farrbconvert.sensorReadings[i];
-      if (reading != -1) {
-        packet_content += (String)reading;
-        packet_content += ",";
+    for (int j=0; j<7; j++) {
```
andycate, line 56: So this means we can have up to 6 floats in the packet?

lib/common_fw/common_fw.cpp
```diff
@@ -36,36 +41,34 @@ bool write_to_SD(std::string message, const char * file_name) {
  * Constructs packet in the following format:
  * {<sensor_ID>,<data1>,<data2>, ...,<dataN>|checksum}
  */
-String make_packet(int id, bool error) {
-  String packet_content = (String)id;
-  packet_content += ",";
+uint8_t make_packet(uint8_t id, bool error) {
```
andycate, line 49: What is `error` for?

lib/common_fw/common_fw.cpp
andycate: Also, these are 8 bit numbers, shouldn't they be 16 bit?

src/E1_coldflow/E1_coldflow.cpp
```diff
@@ -105,26 +100,25 @@ void loop() {
     int i = 0;
     while (RFSerial.available()) {
       command[i] = RFSerial.read();
-      Serial.print(command[i]);
       i++;
     }
+    Serial.write(command, i);
```
andycate: This is just for debugging right? Delete before merge?

src/E1_coldflow/E1_coldflow.cpp
```diff
@@ -105,26 +100,25 @@ void loop() {
...
-      packet = make_packet(valve.id, false);
-      Serial.println(packet);
+      packet_size = make_packet(valve.id, false);
+      Serial.write(packet, packet_size);
       #if SERIAL_INPUT != 1
-        RFSerial.println(packet);
+        RFSerial.write(packet, packet_size);
       #endif
-      write_to_SD(packet.c_str(), file_name);
+      packet[packet_size] = '\0';
```
andycate: What is this for? Is this a delimiter?

andycate: If so, why do we need it? We already have the packet start and packet end bytes

src/E1_coldflow/E1_coldflow.cpp
```diff
@@ -105,26 +100,25 @@ void loop() {
...
+      write_to_SD(packet, file_name);
```
andycate: You don't need to specify how many bytes to write from `packet`? What if `packet` has junk left over at the end of it from a previous packet?

src/E1_coldflow/E1_coldflow.cpp
```diff
@@ -134,12 +128,13 @@ void loop() {
...
+    packet_size = make_packet(sensor->id, false);
+    Serial.write(packet, packet_size);
     #if SERIAL_INPUT != 1
-        RFSerial.println(packet);
-      #endif
-    write_to_SD(packet.c_str(), file_name);
+        RFSerial.write(packet, packet_size);
+    #endif
+    packet[packet_size] = '\0';
+    write_to_SD(packet, file_name);
```
andycate: Same two questions as above

src/E1_coldflow/E1_coldflow.cpp
```diff
@@ -105,26 +100,25 @@ void loop() {
     int i = 0;
     while (RFSerial.available()) {
       command[i] = RFSerial.read();
-      Serial.print(command[i]);
       i++;
     }
```
andycate: Do we not want to check if the first two bytes constitute a start packet command every time a new byte is received?

andycate (commented):
Looks good, just a few questions. I will implement a parser on the ground station side so we can test.

src/E1_coldflow/E1_coldflow.cpp
fazerlicourice7: `write_to_sd` expects input of the format c String which is a character array terminated by the null character `\0`.

src/E1_coldflow/E1_coldflow.cpp
fazerlicourice7: See above. it looks for the `\0` character

lib/common_fw/common_fw.cpp
fazerlicourice7: yes arbitrary. yes they should be 16.

lib/common_fw/common_fw.cpp
fazerlicourice7: ?? Its big endian.
I'm not sure what you're asking.

lib/common_fw/common_fw.cpp
fazerlicourice7: 7

lib/common_fw/common_fw.cpp
fazerlicourice7: error packets. Currently just used for sd init/write errors.

src/E1_coldflow/E1_coldflow.cpp
fazerlicourice7: yeah. I'll incorporate the DEBUG flag.

src/E1_coldflow/E1_coldflow.cpp
fazerlicourice7: I was doing that in the decode packet. But maybe it makes more sense to do that here. I can also then stop when I see the end bytes

fazerlicourice7:
I added the DEBUG flag as a build flag instead of defining it in code. 
Started the String -> std::string refactor

src/E1_coldflow/E1_coldflow.cpp
andycate: :+1: Yeah I think doing it here is better

lib/common_fw/common_fw.cpp
andycate: Yeah, sry. I made this comment when I saw the start_packet and end_packet variables were only 1 byte instead of 2 so I wasn't sure what u were doing here. I understand now tho

lib/common_fw/common_fw.cpp
andycate: Yeet

lib/common_fw/common_fw.cpp
andycate: I see. so if error is true, it sends the id and a zero byte?

src/E1_coldflow/E1_coldflow.cpp
andycate: oh I see. `\0` is the byte `0x00` tho, so what happens if the packet has a zero byte in it?

lib/common_fw/common_fw.cpp
andycate: Ok sg. Is there any reason to choose these more carefully than just arbitrarily?

lib/common_fw/common_fw.cpp
```diff
@@ -77,30 +77,32 @@ uint8_t make_packet(uint8_t id, bool error) {
...
+  int action = _command[i++]; // currently, action is assumed to be a single int.
+  uint16_t checksum = (_command[i++] << 8) & _command[i++];
+  uint16_t _check = Fletcher16(_command+2, 2); // just id and action both of which are ints.
 
-  debug(checksum, DEBUG);
-  debug(_check, DEBUG);
+  int size = sprintf(packet,"%x",checksum);
+  debug(packet, size);
+  size = sprintf(packet,"%x", _check);
+  debug(packet, size);
```
andycate, line 147: Do you mean `_command` instead of `packet` here?

lib/common_fw/common_fw.cpp
```diff
@@ -149,9 +151,16 @@ uint16_t Fletcher16(uint8_t *data, int count) {
...
     Serial.flush();
-  }
+  #endif
+}
+
+void debug(uint8_t *bytes, int count){
+  #ifdef DEBUG
+    Serial.write(bytes, count);
+    Serial.flush();
+  #endif
```
andycate: this might be more useful if you print out the actual hex representation of the packet and not just the packet, because it will be a bunch of junk in the serial console.

lib/common_fw/common_fw.h
```diff
@@ -112,10 +114,11 @@ uint8_t make_packet (uint8_t id, bool error);
 uint16_t Fletcher16 (uint8_t *data, int count);
 void chooseValveById (int id, struct valveInfo *valve, valveInfo valves[], int numValves);
 bool write_to_SD(std::string message, const char * file_name);
-int8_t decode_received_packet(uint8_t *packet, valveInfo *valve, valveInfo valves[], int numValves, int DEBUG);
+int8_t decode_received_packet(uint8_t *packet, valveInfo *valve, valveInfo valves[], int numValves);
```
andycate: do u wanna change `packet` to `_command` here?

platformio.ini
```diff
@@ -12,6 +12,8 @@
 platform = teensy
 board = teensy41
 framework = arduino
+build_unflags = -std=gnu++11
+build_flags = -std=c++17
```
andycate, line 16: what is this for?

andycate (commented):
A few more questions

platformio.ini
fazerlicourice7: specify c++17 as the std to use. some funcs and libs don't exist in older versions

lib/common_fw/common_fw.cpp
fazerlicourice7: no I'm using packet as a temp variable (char *) to hold the result string representation of the checksum, which will then get printed out.

src/E1_coldflow/E1_coldflow.cpp
fazerlicourice7: fixed. It takes a length now

fazerlicourice7:
Updated w/ map for command -> size. 
The internals of the 'decode' still assumes the actual data in the packet is a single element. I think we can change once we actually have a command that has more than one piece of data in it.

taitmate38:
[New packet format](https://trello.com/c/AJpKZWQO/195-new-packet-format)

lib/common_fw/common_fw.cpp
```diff
@@ -9,18 +9,20 @@ SdFat sd;
...
 
 /**
  *
  */
-bool write_to_SD(std::string message, const char * file_name) {
-    sdBuffer->enqueue(message);
+bool write_to_SD(uint8_t *message, uint8_t size, const char * file_name) {
+    sdBuffer->enqueue(message, size);
     if(sdBuffer->length >= 40) {
       if(file.open(file_name, O_RDWR | O_APPEND)) {
```
andycate, line 21: If it's not open, shouldn't this function still pop an item from the front of the queue to prevent it from growing indefinitely?

lib/common_fw/common_fw.cpp
```diff
@@ -32,40 +34,93 @@ bool write_to_SD(std::string message, const char * file_name) {
     return true;
 }
 
+
+/**
+ *
+ */
+bool write_to_SD(std::string message, const char * file_name) {
+    return write_to_SD((uint8_t *)message.c_str(), message.length(), file_name);
+}
```
andycate, line 43: When is this function used? Why don't we just use the one above?

andycate (commented):
Looks done, just two minor questions.

lib/common_fw/common_fw.cpp
fazerlicourice7: Debug. When we're writing constant strings see`setup`

lib/common_fw/common_fw.cpp
fazerlicourice7: No I don't think so. I think we should let it fail so that we know there isn't an additional and then put an SD in.

andycate (approved):
Looks good

fazerlicourice7:
This will not be needed since we're going to switch to eth which will significantly increase bandwidth

PR #41, opened 2021-02-05
Additional auto

daangbruce:


fazerlicourice7 (changes requested):
You didn't merge master in properly. The structure is now `src/E1_coldflow/E1_coldflow.cpp` and the brain_utils has been moved to `lib/general/` or something. Move your changes from brain_utils to the common_fw and move your changes from E1_coldflow to the right file instead of creating a new file

fazerlicourice7:
Also, is this still untested?

PR #43, opened 2021-02-09
Ben/both auto lib

taitmate38:
Cleaned up and moved automated startup & shutdown sequence into `Automation.h`. Tested over multiple weekends @ Bruce's.


taitmate38:
Not sure what's up with the checks - just tested & latest commit successfully builds locally

taitmate38:
[Startup & Shutdown Automation](https://trello.com/c/6olXYIGq/179-startup-shutdown-automation)

taitmate38:
> Not sure what's up with the checks - just tested & latest commit successfully builds locally

Resolved checks - turns out build check couldn't recognize `solenoids.h` dependency & failed due to miscapitalization, whereas with same typo local build would succeed.

src/E1_coldflow/E1_coldflow.cpp
```diff
@@ -139,9 +161,12 @@ void loop() {
     #if SERIAL_INPUT != 1
         RFSerial.println(packet);
       #endif
-    write_to_SD(packet.c_str(), file_name);
+    // write_to_SD(packet.c_str(), file_name);
```
fazerlicourice7: un comment

src/E1_coldflow/E1_coldflow.cpp
```diff
@@ -151,13 +176,23 @@ void loop() {
...
-      break;
+      #if NO_ADC == 1
+        break;
+      #else
+        Thermocouple::Analog::readTemperatureData(farrbconvert.sensorReadings);
+        farrbconvert.sensorReadings[1] = tempController::controlTemp(farrbconvert.sensorReadings[0]);
+        farrbconvert.sensorReadings[2] = -1;
+        break;
+      #endif
+
```
fazerlicourice7: can you add some comments wherever this shows saying this a temp quick fix. and add `TODO (@fazerlicourice7 ): do it properly` 
I'm already working on it.

src/E1_coldflow/E1_coldflow.cpp
```diff
@@ -118,7 +120,27 @@ void loop() {
       #if SERIAL_INPUT != 1
         RFSerial.println(packet);
       #endif
-      write_to_SD(packet.c_str(), file_name);
+      // write_to_SD(packet.c_str(), file_name);
```
fazerlicourice7: un comment

src/E1_coldflow/E1_coldflow.cpp
```diff
@@ -62,7 +64,7 @@ void setup() {
   debug("Sensor IDs:", DEBUG);
   debug(String(sensors[0].name), DEBUG);
 
-  debug("Starting SD", DEBUG);
+  // debug("Starting SD", DEBUG);
```
fazerlicourice7: un comment

lib/Automation/Automation.h
```diff
@@ -0,0 +1,219 @@
+/*
+  Automation.h - A c++ library including automated sequences triggered by
+  commands or particular events (e.g. end of flow, overpressurization)
+*/
+
+
```
fazerlicourice7: I recall that we came to conclusion that we were going to keep the libraries a consistent format (classes w/ .h interface definition, and .cpp source). 
Low key I don't care that much. But thought I would remind you anway. 
(I've been doing this whenever I had reason to edit a specific file)

fazerlicourice7 (commented):
minor things

lib/Automation/Automation.h
taitmate38: Oh yeah good point - I'll update it to that format for consistency.

lib/Automation/Automation.cpp
```diff
@@ -0,0 +1,198 @@
...
+    3 - Between open Prop Main and close Main Valve Arm
+  */
+  int _startupDelays[3] = {1000, 0 ,1000};
+
+  /* Delays during shutdown sequence:
+    1 - Between arm 2-way, close high pressure, close Prop and close LOX
+    2 - Between close LOX and closing Arming valve & opening Gems vent
+  */
+  int _shutdownDelays[2] = {0, 750};
+
```
fazerlicourice7, line 37: Shouldn't these be a config file sorta thing.

lib/Automation/Automation.cpp
```diff
@@ -0,0 +1,197 @@
...
+        Solenoids::openLOX();
+        Solenoids::openPropane();
+        Solenoids::getAllStates(data);
+        _startupPhase++;
+
+    } else if (_startupPhase == 2) {
+      
+        // after a delay open Prop main valve
+        // Solenoids::openPropane();
+        Solenoids::getAllStates(data);
```
fazerlicourice7, line 126: So this sequence will be different for coldflow, waterflow, hotfire?

in that case, I think we need to think more about how we should implement it so that those changes are easy to make. 
more atomic functions: each of the contents of the `if`s should be its own function. And we should be able to specify the order of doing them in the config file.

lib/solenoids/solenoids.cpp
```diff
@@ -0,0 +1,322 @@
...
+  solenoids.h - A c++ library to interface with all of the solenoids on our rocket.
+  Created by Vamshi Balanaga, Sept 15, 2020.
+*/
+
+#include "solenoids.h"
+
+using namespace std;
+
+namespace Solenoids {
+
```
fazerlicourice7, line 11: Can you just make this one a class

fazerlicourice7 (changes requested):
I'm primarily concerned about the modularity of the automation sequence.

taitmate38:
Changes made to make generalization more general. Significant overhaul of approach to automation, now on separate branch: `ben/auto_and_detect`, so closing this PR

PR #45, opened 2021-02-10
Neha/pcktcounter

nehading:
followed Vamshi's instructions to add packet counter to common_fw, used in E1_coldflow to increment when sending packets via RF & added as sensor reading
- fixed build errors from earlier
- replaced broken file.write() line with code from ofstream library
- would not link since _open system call was not defined by OS for some reason --> provided stub empty implementation of _open so it fails safely for now

### DIFF
--- a/lib/common_fw/common_fw.cpp
+++ b/lib/common_fw/common_fw.cpp
@@ -15,7 +15,6 @@ union floatArrToBytes farrbconvert;
  *
  */
 bool write_to_SD(std::string message, const char * file_name) {
-    using namespace std;
     sdBuffer->enqueue(message);
     if(sdBuffer->length >= 40) {
       if(file.open(file_name, O_RDWR | O_APPEND)) {
--- a/lib/common_fw/common_fw.cpp
+++ b/lib/common_fw/common_fw.cpp
@@ -9,6 +9,7 @@ SdFat sd;
 File file;
 struct Queue *sdBuffer;
 union floatArrToBytes farrbconvert;
+int packetCounter = 0;
 
 /**
  *
@@ -106,6 +107,15 @@ int decode_received_packet(String packet, valveInfo *valve, valveInfo valves[], 
   } else {
     return -1;
   }
+}
+
+void readPacketCounter(float *data) {
+    data[0] = packetCounter;
+    data[1] = -1;
+  }
+
+void incrementPacketCounter() {
+    packetCounter+=1;
 }
 
 /**
--- a/lib/common_fw/common_fw.cpp
+++ b/lib/common_fw/common_fw.cpp
@@ -7,9 +7,9 @@
 
 SdFat sd;
 File file;
+int packetCounter = 0;
 struct Queue *sdBuffer;
 union floatArrToBytes farrbconvert;
-int packetCounter = 0;
 
 /**
  *
@@ -68,7 +68,7 @@ String make_packet(int id, bool error) {
   }
   packet_content += check_;
   String packet = "{" + packet_content + "}";
-
+  incrementPacketCounter();
   return packet;
 }
 
--- a/lib/common_fw/common_fw.cpp
+++ b/lib/common_fw/common_fw.cpp
@@ -18,16 +18,14 @@ bool write_to_SD(std::string message, const char * file_name) {
     using namespace std;
     sdBuffer->enqueue(message);
     if(sdBuffer->length >= 40) {
-      fstream currfile;
-      currfile.open(file_name, fstream::out | fstream::app);
-      if(currfile.is_open()) {
+      if(file.open(file_name, O_RDWR | O_APPEND)) {
         int initialLength = sdBuffer->length;
         for(int i = 0; i < initialLength; i++) {
           char *msg = sdBuffer->dequeue();
-          currfile.write(msg, sizeof(msg)); 
+          file.write(msg, sizeof(msg)); 
           free(msg);
         }
-        currfile.close();
+        file.close();
         return true;
       } else {        //If the file didn't open
           return false;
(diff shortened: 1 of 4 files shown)


nehading:
[Additional Telemetry](https://trello.com/c/65rseMSX/184-additional-telemetry)

src/E1_coldflow/E1_coldflow.cpp
```diff
@@ -68,6 +68,7 @@ void setup() {
   if (!res) {
     packet = make_packet(101, true);
     RFSerial.println(packet);
+    incrementPacketCounter();
```
fazerlicourice7: Notice how to increment is always called along with make packet? I think we should just put the increment function inside `make_packet`

lib/common_fw/common_fw.cpp
```diff
@@ -9,24 +9,28 @@ SdFat sd;
...
 /**
  *
  */
 bool write_to_SD(std::string message, const char * file_name) {
+    using namespace std;
     sdBuffer->enqueue(message);
     if(sdBuffer->length >= 40) {
-      if(file.open(file_name, O_RDWR | O_APPEND)) {
+      fstream currfile;
+      currfile.open(file_name, fstream::out | fstream::app);
```
fazerlicourice7: Not sure I understand why you did this. We have `file` defined in config

lib/common_fw/common_fw.cpp
```diff
@@ -9,24 +9,28 @@ SdFat sd;
 File file;
 struct Queue *sdBuffer;
 union floatArrToBytes farrbconvert;
+int packetCounter = 0;
```
fazerlicourice7: I would put this in header file

lib/common_fw/common_fw.cpp
nehading: apparently file.write() doesn't exist in c++ - have to use fsteam to write to files (so I opened the file w fstream to write to it)

lib/common_fw/common_fw.cpp
nehading: for some weird reason whenever I define something in the .h file and then build - it errors and says that variable is defined twice - do you know why this happens @fazerlicourice7

lib/common_fw/common_fw.cpp
nehading: or is this an arduino thing

lib/common_fw/common_fw.cpp
fazerlicourice7: define it only in .h and not in .cpp

lib/common_fw/common_fw.cpp
fazerlicourice7: It's an arduino thing. it should work if you're using pio to compile

lib/common_fw/common_fw.cpp
nehading: yeah it errors when i only defined it in .h

src/E1_coldflow/config.h
```diff
@@ -71,6 +72,9 @@ namespace config {
     sensors[1] = {"All Pressure",  FLIGHT_BRAIN_ADDR, 1, 1};
     sensors[2] = {"Battery Stats", FLIGHT_BRAIN_ADDR, 2, 3};
     sensors[3] = {"Aux temp",      FLIGHT_BRAIN_ADDR, 4, 1};
+    sensors[5] = {"Number Packets Sent", FLIGHT_BRAIN_ADDR, 5, 1};
```
fazerlicourice7: make the last number here 10, and add the packet id and details to this sheet https://docs.google.com/spreadsheets/d/1xNmjrb5a28FfVg-MG9jUBRiPyfdkjo7HiqPWP41N_IQ/edit#gid=998950668 in the `packet-id` tab

lib/common_fw/common_fw.cpp
fazerlicourice7: can you push the exact code that causes that error along w/ a screen shot of the full error message

nehading: so on stack overflow it says variables have to be declared w/ extern in the .h file and defined/initialized in the .cpp file otherwise you get the multiple definitions error ([https://stackoverflow.com/questions/38942013/declaring-variables-in-header-files-c](url) (followed the first answer) - also sorry for all the back & forth

lib/common_fw/common_fw.cpp
```diff
@@ -7,26 +7,28 @@
...
 File file;
+int packetCounter = 0;
 struct Queue *sdBuffer;
 union floatArrToBytes farrbconvert;
 
 /**
  *
  */
 bool write_to_SD(std::string message, const char * file_name) {
+    using namespace std;
```
fazerlicourice7: unnecessary

src/E1_coldflow/config.h
```diff
@@ -71,6 +72,9 @@ namespace config {
     sensors[1] = {"All Pressure",  FLIGHT_BRAIN_ADDR, 1, 1};
     sensors[2] = {"Battery Stats", FLIGHT_BRAIN_ADDR, 2, 3};
     sensors[3] = {"Aux temp",      FLIGHT_BRAIN_ADDR, 4, 1};
+    sensors[5] = {"Number Packets Sent", FLIGHT_BRAIN_ADDR, 5, 10};
+
+
```
fazerlicourice7: Let's not add random empty lines unneccesarily. one is fine.

fazerlicourice7 (commented):
couple tiny things, but in general looks good.

lib/common_fw/common_fw.cpp
fazerlicourice7: shit you're right. I definitely do that too :P my b.

fazerlicourice7 (approved):

PR #49, opened 2021-02-21
Vamshi/command processing redo

fazerlicourice7:
kinda just whipped this together. Not even sure if this is the best architecture for this. Probably requires some cleanup.
Needs to be tested on the hardware. Would be great if someone could take a look and test it out.


lib/common_fw/common_fw.cpp
```diff
@@ -68,47 +68,93 @@ String make_packet(int id, bool error) {
...
+// }
 
 /*
  * Decodes a packet sent from ground station in the following format:
  * {<valve_ID>,<open(1) or close(0)|checksum}
  * Populated the fields of the valve and returns the action to be taken
  * This is a pretty beefy function; can we split this up
  */
-int decode_received_packet(String packet, valveInfo *valve, valveInfo valves[], int numValves) {
+uint8_t parseCommand(String packet) {
```
nehading: should it be signed since returns -1?

lib/communications/ActuatorArray.cpp
```diff
@@ -0,0 +1,56 @@
+
+#include "actuator.h"
+#include <cstdlib>
+#include <Arduino.h>
+
+
+uint8_t binary_search(uint8_t *array, uint8_t start, uint8_t end, uint8_t val);
```
nehading: I think this should also be signed since returns -1

lib/tempController/tempController.h
```diff
@@ -25,9 +27,44 @@ class TempController {
 
     PID *controller = new PID(255, 0, k_p, k_i, k_d);
 
+  protected:
+    bool humanOverride = false;
+    uint16_t humanSpecifiedValue = 0;
```
nehading: maybe set the default value to 300 since corresponds w default false value of humanOverride

src/E1_coldflow/E1_coldflow.cpp
```diff
@@ -109,10 +105,10 @@ void loop() {
     }
 
     debug(String(command), DEBUG);
-    int action = decode_received_packet(String(command), &valve, valves, numValves);
-    if (action != -1) {
-      take_action(&valve, action);
-      packet = make_packet(valve.id, false);
+    uint8_t id = parseCommand(String(command));
```
nehading: should it be signed?

nehading (commented):
closely looked through the code that I could make sense of and it looks perfect to me other than the small thing i pointed out

fazerlicourice7:
[Generalize commands library](https://trello.com/c/WRNYA5ec/171-generalize-commands-library)

fazerlicourice7:
So, This is a huge PR that include a rearchitecture of the command processing. Using an `actuator` interface and HeaterActuators, SolenoidActuator and AutomationActuators that all extend the interface. 

this also includes modifications to the SD queue: It has moved  from a linked list implementation to a full static array implementation. Looks like this was the cause of the crashing. and It's looking very good right now. 

Also included is several different dynamic -> static changes (object arrays vs object pointers + malloc basically)

A lot of changes are also auto formatting, which is why some files have so many changes.

lib/Automation/Automation.h
```diff
@@ -108,16 +108,55 @@ extern struct autoEventList* _eventList;
...
+      Automation::flowConfirmation(data);
+    }
+
+  private:
+    func_t startSequence;
+    func_t endSequence;
+
+};
+
+extern AutomationSequenceActuator fullFlow;
```
taitmate38: Is there a way we can do it where Actuators aren't defined as a part of the library? Maybe that makes sense for libraries where there are a defined number of actuators, but I feel like with the Automation what AutomationSequenceActuators we're actually should be something we determine at the config level, since it could change

fazerlicourice7:
Nolan's running it over night, and that will be the final crashing verification.

And we tested a subset of commands to make sure it's working (everything we teste dworked, but we didn't test everythign).

src/E1_coldflow/E1_coldflow.cpp
```diff
@@ -23,22 +22,19 @@
...
-char command[50]; //input command from GS
+char command[75]; //input command from GS
 
 /*
     Stores how often we should be requesting data from each sensor.
 */
 int sensor_checks[numSensors][2];
 
-valveInfo valve;
+//valveInfo valve;
```
taitmate38: can delete?

src/E1_coldflow/config.h
```diff
@@ -1,27 +1,29 @@
...
 #include "common_fw.h"
 #include <ADS1219.h>
 #include "Automation.h"
 
-
 #define FLIGHT_BRAIN_ADDR 0x00
 #define DEBUG 1
 
-std::string str_file_name = "E1_speed_test_results.txt";
+std::string str_file_name = "E1_coldflow.txt";
```
taitmate38, line 12: is a better/dynamic naming convention being done on a different PR?

src/E1_coldflow/E1_coldflow.cpp
```diff
@@ -106,15 +101,14 @@ void loop() {
...
       i++;
     }
 
-    debug(String(command), DEBUG);
-    int action = decode_received_packet(String(command), &valve, valves, numValves);
-    if (action != -1) {
-      take_action(&valve, action);
-      packet = make_packet(valve.id, false);
+    debug(String(command));
+    int8_t id = parseCommand(String(command));
```
taitmate38: I feel like parseCommand should be made to not actually take the action - that way the parsing of the bytes is separate from what you actually do with them, which makes it clearer & more flexible.

lib/common_fw/common_fw.cpp
```diff
@@ -68,47 +73,99 @@ String make_packet(int id, bool error) {
...
+  debug(String(data));
 
   int _check = (int)Fletcher16((uint8_t *) data, count);
-  Serial.println(_check);
   if (_check == checksum) {
-    Serial.println("Checksum correct, taking action");
-    chooseValveById(valve_id, valve, valves, numValves);
-    return action;
+    debug("Checksum correct, taking action");
+    tmpActuator = actuators.get(actuator_id); //chooseValveById(valve_id, valve, valves, numValves);
```
taitmate38: Feel like this section that actually takes the action/confirmation should be moved out of `parseCommand` and into a separate function (`takeAction`?), to keep the parsing separate from the actual action that is being taken based off the incoming packet. Would also make it so we could interpret the `id` & `command_data` as something that is separate than an actuator.

src/E1_coldflow/config.h
```diff
@@ -83,20 +84,42 @@ namespace config {
...
-    valves[4] = {"Propane 5 Way", 24, &(Solenoids::openPropane), &(Solenoids::closePropane), &(Solenoids::getAllStates)};
-    valves[5] = {"Propane GEMS", 25, &(Solenoids::ventPropaneGems), &(Solenoids::closePropaneGems), &(Solenoids::getAllStates)};
-    valves[6] = {"High Pressure Solenoid", 26, &(Solenoids::activateHighPressureSolenoid), &(Solenoids::deactivateHighPressureSolenoid), &(Solenoids::getAllStates)};
-    valves[7] = {"Arm Rocket", 27, &(Solenoids::armAll), &(Solenoids::disarmAll), &(Solenoids::getAllStates)};
-    valves[8] = {"Launch Rocket", 28, &(Solenoids::LAUNCH), &(Solenoids::endBurn), &(Solenoids::getAllStates)};
-    valves[9] = {"Perform Flow", 29, &(Automation::beginBothFlow), &(Automation::endBothFlow), &(Automation::flowConfirmation)};
-    valves[10] = {"Perform LOX Flow", 30, &(Automation::beginLoxFlow), &(Automation::endLoxFlow), &(Automation::flowConfirmation)};
-
+    debug("Initializing actuators");
+    Solenoids::lox_2.setId(20);
```
taitmate38: given that the # of valves themselves that we are going to use will not change, maybe some of this can be done in the solenoids init? Just like how the pin mappings are all just passed in, maybe the valve IDs could be too.

src/E1_waterflow/E1_waterflow.cpp
```diff
@@ -105,23 +99,23 @@ void loop() {
...
 
-    debug(String(command), DEBUG);
-    int action = decode_received_packet(String(command), &valve, valves, numValves);
-    if (action != -1) {
-      take_action(&valve, action);
-      packet = make_packet(valve.id, false);
+    debug(String(command));
+    int8_t id = parseCommand(String(command));
+    if (id != -1) {
+      //take_action(&valve, action);
```
taitmate38: delete

src/E1_waterflow/E1_waterflow.cpp
```diff
@@ -105,23 +99,23 @@ void loop() {
...
       i++;
     }
 
-    debug(String(command), DEBUG);
-    int action = decode_received_packet(String(command), &valve, valves, numValves);
-    if (action != -1) {
-      take_action(&valve, action);
-      packet = make_packet(valve.id, false);
+    debug(String(command));
+    int8_t id = parseCommand(String(command));
```
taitmate38: Same comment as on E1_coldflow.cpp - think it would be clearer if parseCommand only parsed the command, and the logic to take the action was somewhere else. Could allow certain IDs to be used to get an actuator, but then maybe have others just set a flag or change a parameter, without necessarily needing a whole actuator.

lib/common_fw/common_fw.cpp
```diff
@@ -68,47 +73,99 @@ String make_packet(int id, bool error) {
...
     return -1;
   }
-  int action = packet.substring(data_start_index + 1,data_end_index).toInt();
+  debug(String(data_end_index));
+  debug(packet.substring(data_start_index + 1,data_end_index).c_str());
+  char data_string[data_end_index - data_start_index + 1];
+  packet.substring(data_start_index + 1,data_end_index).toCharArray(data_string, data_end_index - data_start_index + 1);
+  char *tmp;
+  uint8_t i = 0;
+  float command_data[7];
```
taitmate38: should have the maximum # of floats in a command as a constant/#define somewhere else

lib/common_fw/common_fw.cpp
```diff
@@ -68,47 +73,99 @@ String make_packet(int id, bool error) {
...
+    tmpActuator = actuators.get(actuator_id); //chooseValveById(valve_id, valve, valves, numValves);
+    tmpActuator->parseCommand(command_data);
+    tmpActuator->confirmation(farrbconvert.sensorReadings);
+    return tmpActuator->ID();
   } else {
     return -1;
   }
 }
 
+
```
taitmate38: delete

lib/common_fw/common_fw.cpp
```diff
@@ -8,28 +8,33 @@
...
-          free(msg);
-        }
-        file.close();
-        return true;
-      } else {        //If the file didn't open
-          return false;
-      }
-    }
+    sdBuffer->dequeue(buffer);
+    // if(sdBuffer->length >= 40) {
```
taitmate38: delete

lib/communications/ActuatorArray.cpp
```diff
@@ -0,0 +1,80 @@
...
+#include "actuator.h"
+#include <cstdlib>
+#include <Arduino.h>
+
+
+int8_t binary_search(uint8_t *array, uint8_t start, uint8_t end, uint8_t val);
+int8_t sequential_search(uint8_t *array, uint8_t size, uint8_t val);
+
+ActuatorArray::ActuatorArray(const uint8_t size) {
+  _backingStore = (Actuator **) std::malloc(size * sizeof(Actuator *));
```
taitmate38: if we have a set `size` number of Actuators, can't this just be a static array that is created in the highest level stack frame and passed down as needed? To avoid needing to use malloc, since I imagine this thing is intended to be created once at the beginning & used as-is for the rest of the execution of the code

lib/solenoids/solenoids.cpp
```diff
@@ -244,6 +246,7 @@ namespace Solenoids {
   }
 
   int openLOX() {
+    Serial.println("open lox");
```
taitmate38: remove or set to debug message?

lib/ADS1219/native/mock_ADS1219.cpp
```diff
@@ -1,12 +1,18 @@
...
+ADS1219::ADS1219() {
+
+}
+
+void ADS1219::init(int drdy, uint8_t addr, TwoWire *wire) {
+  // do nothing.
+}
+
 long ADS1219::getData(uint8_t conf) {
   // do nothing.
```
taitmate38: would be nice if there were an option to make this produce fake/random data, instead of just doing nothing. Would mean can still test things like connection to the dashboard (and could even become a tool itself for verifying the dashboard is graphing correctly w/o needing the actual PCB).

taitmate38 (changes requested):

fazerlicourice7:
So, I just got up, and the code is still running. 
But,it's very, very,very slow packets at ~1Hz.......
@nlautrette , do you see this in the morning as well? 
I'm going to play around w/ this over the weekend to see what could possibly be causing this. SD? 
Have we seen this before?

fazerlicourice7:
See vid: https://drive.google.com/file/d/1LbeQppT67xiXMzRx3tqEi-GP_11i5UIz/view?usp=sharing

taitmate38:
> So, I just got up, and the code is still running.
> But,it's very, very,very slow packets at ~1Hz.......
> @nlautrette , do you see this in the morning as well?
> I'm going to play around w/ this over the weekend to see what could possibly be causing this. SD?
> Have we seen this before?

You mean the code was running faster and then slowed down over time?
Either way, both E1_waterflow.cpp & E1_coldflow.cpp both have delay statements in the main loop (10ms & 50ms respectively). That's probably not your issue, but won't be helping in any way.

fazerlicourice7:
yeah it was blazing fast when I went to bed, and got to 1Hz in the morning.... def not delay related.

src/E1_coldflow/config.h
fazerlicourice7: yeah.

src/E1_coldflow/config.h
fazerlicourice7: yeah passing in the ids sounds good. I just didn't want to hardcode ids outside of the config file

src/E1_waterflow/E1_waterflow.cpp
fazerlicourice7: the "actuator" is an abstract class. It can be anything we want it to be doesn't need to be a solenoid or heater. Actions like setting flags and changing parameters can be done w/in the context of an "actuator". Essentially they are virtual actuators. This prevents you from having divergences in the code. EVerything is handled w/ one case.

lib/common_fw/common_fw.cpp
fazerlicourice7: See below comment. Actuator is a general object. Any command processing should be done in the context of an actuator. 
I can rename actuator to something more general if that helps?

lib/common_fw/common_fw.h
```diff
@@ -120,31 +49,38 @@ struct Queue {
...
   }
 };
 
+const uint8_t maxReadings = 9;
+
 /*
  * Data structure to allow the conversion of bytes to floats and vice versa.
  */
 union floatArrToBytes {
   char buffer[36];
```
taitmate38: this could also be generalized as `char buffer[4*maxReadings]`, right?

taitmate38 (dismissed):

nlautrette (approved):

nlautrette:
Looks good just approved

PR #54, opened 2021-03-29
Working draft of lad-4 flight code

nlautrette:


lib/Barometer/Barometer.cpp
```diff
@@ -0,0 +1,65 @@
+/*
+  IMU.cpp - A c++ library to interface with the Arduino BMP3XX Barometer.
+  Created by Nolan Lautrette, March 21, 2021.
+*/
```
fazerlicourice7: fix comment

lib/GPS/GPS.cpp
```diff
@@ -0,0 +1,81 @@
...
+/**
+ * Define this in advance. Need to be agreed on by everyone.
+ */
+void GPS::readAuxilliaryData(float *data) {
+  _gps.read();
+  data[0] = _gps.altitude;
+  data[1] = _gps.speed;
+  data[2] = _gps.angle;
+  data[3] = _gps.satellites;
+  data[4] = -1;
```
fazerlicourice7: change this order based on packet def. Add fix data (is that just num satellites? or was there extra). 
And update packet def to include angle

lib/apogeeDetection/apogeeDetection.cpp
```diff
@@ -5,35 +5,40 @@
...
+		_R = new MatrixXd(m, m);
 		*_R << altitudeVar , 0,
 		 			0,     accelVar;
 
-		_z = new VectorXd(2);
+		_z = new VectorXd(m);
 
-		kalmanfilter = new Kalman(3, *_F, 2, 1, *_H, MatrixXd::Zero(3,1), *_Q, *_R);
+		_x = new VectorXd(n);
+		*_x << initAlt, initAcc, 0;
```
fazerlicourice7: this should be `initAlt, 0, initAcc;`

platformio.ini
```diff
@@ -24,16 +24,16 @@ lib_deps =
 
 [env:E1_coldflow]
 src_filter = -<*> +<E1_coldflow/>
-build_flags = -DHAL=teensy ; -DDEBUG
```
fazerlicourice7: undo these changes. we want to be able to easily uncomment, or comment out the DEBUG flag

src/LAD4_1/config.h
```diff
@@ -1,72 +1,34 @@
...
-const uint8_t numSensors = 4;
-struct sensorInfo **sensors;
+const uint8_t numSensors = 7;
+sensorInfo *sensors;
 
 const float batteryMonitorShuntR = 0.002; // ohms
 const float batteryMonitorMaxExpectedCurrent = 10; // amps
 
-#define DROGUE_PIN 9
-#define MAIN_CHUTE_PIN 10
```
fazerlicourice7, line 33: this should stay. make it a `const uint8_t` though.

src/LAD4_1/config.h
```diff
@@ -1,72 +1,34 @@
...
+    sensors[0] = {"IMU Acceleration",   FLIGHT_BRAIN_ADDR, 14, 1};
+    sensors[1] = {"IMU Orientation", FLIGHT_BRAIN_ADDR, 15, 1};
+    sensors[2] = {"Barometer", FLIGHT_BRAIN_ADDR, 13, 1};
+    sensors[3] = {"Battery Stats", FLIGHT_BRAIN_ADDR, 2, 3};
+    sensors[4] = {"GPS Lat Long",  FLIGHT_BRAIN_ADDR, 11, 3};
+    sensors[5] = {"GPS AUX",      FLIGHT_BRAIN_ADDR, 12, 3};
+    sensors[6] = {"Number Packets Sent", FLIGHT_BRAIN_ADDR, 5, 10};
 
-    pinMode(DROGUE_PIN, OUTPUT);
-    pinMode(MAIN_CHUTE_PIN, OUTPUT);
```
fazerlicourice7: This stays

src/LAD4_1/config.h
```diff
@@ -1,72 +1,34 @@
...
+    //debug("Initializing sensors", DEBUG);
+    sensors = new sensorInfo[numSensors];
+    // the ordering in this array defines order of operation, not id
+    sensors[0] = {"IMU Acceleration",   FLIGHT_BRAIN_ADDR, 14, 1};
+    sensors[1] = {"IMU Orientation", FLIGHT_BRAIN_ADDR, 15, 1};
+    sensors[2] = {"Barometer", FLIGHT_BRAIN_ADDR, 13, 1};
+    sensors[3] = {"Battery Stats", FLIGHT_BRAIN_ADDR, 2, 3};
+    sensors[4] = {"GPS Lat Long",  FLIGHT_BRAIN_ADDR, 11, 3};
+    sensors[5] = {"GPS AUX",      FLIGHT_BRAIN_ADDR, 12, 3};
+    sensors[6] = {"Number Packets Sent", FLIGHT_BRAIN_ADDR, 5, 10};
```
fazerlicourice7, line 45: you will need to add a `sensor` that checks apogee. We Also need to add a sensor for main chute deployment at 300m or something.

src/LAD4_1/lad4_1.cpp
```diff
@@ -1,5 +1,176 @@
...
+
+void sensorReadFunc(int id);
+
+IMU _imu;
+Barometer bmp;
+GPS gps(GPS_Serial);
+ApogeeDetection detector;
+
+double acc_z;
+double altitude;
```
fazerlicourice7: why are these two here. we shouldn't need any of this here. all of that should go inside the apogee detection library.

src/LAD4_1/lad4_1.cpp
```diff
@@ -1,5 +1,176 @@
...
+
+  bmp.init(&Wire);
+
+  gps.init();
+
+  // Initializes initial state _x in Kalman filter to the alt and acc of the rocket at setup
+  sensorReadFunc(0);
+  acc_z = farrbconvert.sensorReadings[2];
+  sensorReadFunc(4);
+  altitude = farrbconvert.sensorReadings[0];
```
fazerlicourice7: there is no case 0 or 4 in `sensorReadFunc`

src/LAD4_1/lad4_1.cpp
```diff
@@ -1,5 +1,176 @@
...
+
+  double altVar = 0.5;
+  double accVar = 0.5;
```
fazerlicourice7: these two go in config as `const double`

src/LAD4_1/lad4_1.cpp
```diff
@@ -1,5 +1,176 @@
...
+  detector = ApogeeDetection(20e-3, altVar, accVar, altitude, acc_z);
```
fazerlicourice7: If we can't construct this statically, let's modify the apogee detection library to be like the IMU and barometer libraries above w/ a default constructor and a separate init func that we can pass all of this to.
20e-3 should also be a variable defined in config.

src/LAD4_1/lad4_1.cpp
```diff
@@ -1,5 +1,176 @@
...
+    sensor = &sensors[j];
+    sensorReadFunc(sensor->id);
+    if(sensor->id == 0) {
+      acc_z = farrbconvert.sensorReadings[0];
+    }
+    else if(sensor->id == 2) {
+      altitude = farrbconvert.sensorReadings[2];
+      if(detector.atApogee(altitude, acc_z)) {
+        //deloy chute
+      }
```
fazerlicourice7: You need to define sensors in config. and move any of this logic to the `sensorReadFunc` defined below.
Ig, you'll need the altitude and acc_z variables you defined above. so ignore that comment.

src/LAD4_1/lad4_1.cpp
```diff
@@ -1,5 +1,176 @@
...
+void sensorReadFunc(int id) {
+  switch (id) {
+    case 14:
+      _imu.readAccelerationData(farrbconvert.sensorReadings);
+      break;
+    case 15:
+      _imu.readOrientationData(farrbconvert.sensorReadings);
+      break;
+    case 13:
+      bmp.readAllData(farrbconvert.sensorReadings);
```
fazerlicourice7: update altitude data here.

src/LAD4_1/lad4_1.cpp
```diff
@@ -1,5 +1,176 @@
...
+    write_to_SD(packet.c_str(), file_name);
+  }
+  delay(10);
+
+}
+
+void sensorReadFunc(int id) {
+  switch (id) {
+    case 14:
+      _imu.readAccelerationData(farrbconvert.sensorReadings);
```
fazerlicourice7: update accel data here.

src/LAD4_1/lad4_1.cpp
```diff
@@ -1,5 +1,176 @@
...
+        RFSerial.println(packet);
+    #endif
+    write_to_SD(packet.c_str(), file_name);
+  }
+  delay(10);
+
+}
+
+void sensorReadFunc(int id) {
+  switch (id) {
```
fazerlicourice7, line 130: update the switch to be in case order 2, 5, 11, 12, 13, 14, 15....

fazerlicourice7 (changes requested):

src/LAD4_1/config.h
nlautrette: @ssalcedo00

lib/apogeeDetection/apogeeDetection.cpp
nlautrette: @ssalcedo00

src/LAD4_1/lad4_1.cpp
nlautrette: @ssalcedo00

platformio.ini
nlautrette: is this supposed to be grayed out @fazerlicourice7

fazerlicourice7: leave the `-DDEBUG` as a comment (the `;` is the comment identifier in`.ini` files)

salsquared: is this resolved?

lib/recovery/recovery.h
```diff
@@ -19,8 +20,8 @@ namespace Recovery {
     drogue_chute_state = 0;
     main_chute_state = 0;
 
-    digitalWrite(drogue_pin, drogue_chute_state);
-    digitalWrite(main_pin, main_chute_state);
+    digitalWrite(DROGUE_PIN, drogue_chute_state);
+    digitalWrite(MAIN_CHUTE_PIN, main_chute_state);
```
fazerlicourice7: nope. update the init to take in two pin numbers, set the value of `drogue_pin` and `main_pin` to the values that were passed in . Use those two vars locally

lib/apogeeDetection/apogeeDetection.cpp
```diff
@@ -57,5 +65,16 @@ bool ApogeeDetection::atApogee(double altitude, double accel_z) {
 	return false;
 }
 
+void ApogeeDetection::updateAlt(float* data) {
+	altitude = data[0];
+}
+
+void ApogeeDetection::updateAcc(float* data) {
+	acc_z = data[2];
+}
```
fazerlicourice7: nice.
send in just a single float though. 
it should be called like `updateAlt(farrbconvert.sensorReadings[0])`

src/LAD4_1/lad4_1.cpp
```diff
@@ -1,5 +1,181 @@
...
+  debug("Initializing Libraries");
+  batteryMonitor::init(&Wire, batteryMonitorShuntR, batteryMonitorMaxExpectedCurrent);
+  _imu.init(&Wire);
+  bmp.init(&Wire);
+  gps.init();
+
+  bmp.readAllData(farrbconvert.sensorReadings);
+  initAcc_z = farrbconvert.sensorReadings[2];
+  _imu.readAccelerationData(farrbconvert.sensorReadings);
+  initAlt = farrbconvert.sensorReadings[0];
```
fazerlicourice7: check this order.

src/LAD4_1/lad4_1.cpp
```diff
@@ -1,5 +1,181 @@
...
+  if(290 <= alt && alt <= 310) 
+    return true;
+  else 
+    return false;
+}
+
+void recoveryPacket() {
+  Recovery::getAllStates(farrbconvert.sensorReadings);
+  packet = make_packet(10, false);
+  Serial.println(packet);
```
fazerlicourice7, line 185: also send to rf, and write to sd

src/LAD4_1/config.h
```diff
@@ -1,70 +1,47 @@
...
```
fazerlicourice7, line 45: add sensor for recovery pack

src/LAD4_1/lad4_1.cpp
```diff
@@ -1,5 +1,181 @@
...
+        Recovery::releaseMainChute();
+        recoveryPacket();
+      }
+      break;
+    case 14:
+      _imu.readAccelerationData(farrbconvert.sensorReadings);
+      if(detector.atApogee()) {
+        passedApogee = true;
+        Recovery::releaseDrogueChute();
+        recoveryPacket();
```
fazerlicourice7, line 161: you'll need to close the chute after a certain delay go with 10ms for now, Andy is testing, we may update that number.

src/LAD4_1/lad4_1.cpp
```diff
@@ -1,5 +1,181 @@
...
+      bmp.readAllData(farrbconvert.sensorReadings);
+      detector.updateAlt(farrbconvert.sensorReadings);
+      if(passedApogee && aroundMainDeploy()) {
+        Recovery::releaseMainChute();
+        recoveryPacket();
+      }
+      break;
+    case 14:
+      _imu.readAccelerationData(farrbconvert.sensorReadings);
+      if(detector.atApogee()) {
```
fazerlicourice7: remember we only need to do this if we're past main engine cut off. You added that to the python lib, let's add that here too

fazerlicourice7 (changes requested):
looking much better. Still some work before it's ready. 
Last stretch!

platformio.ini
fazerlicourice7: yeah

src/LAD4_1/config.h
salsquared: i thought we werent gonna add a sensor for recovery

lib/GPS/GPS.cpp
```diff
@@ -0,0 +1,82 @@
...
+GPS::GPS(SPIClass *theSPI, int8_t cspin) {
+  _gps = Adafruit_GPS(theSPI, cspin);
+  commMethod = 4;
+  init();
+}
+
+void GPS::init() {
+    _gps.begin(9600);
+
+    _gps.sendCommand(PMTK_SET_NMEA_OUTPUT_RMCONLY);
```
fazerlicourice7: Change this to `PMTK_SET_NMEA__OUTPUT_RMCGGA`

src/LAD4_1/config.h
fazerlicourice7: we should send it periodically (very slowly) 
but also send it right after you do the recovery commands

salsquared: every 10, 20 cycles good?

src/LAD4_1/lad4_1.cpp
salsquared: what is there to close lol we're using black powder. or is it gonna be sending a signal until i close it?

fazerlicourice7:
make delay like 500 ms for chute

PR #60, opened 2021-04-01
Working version of lad4 flight code

nlautrette:
fixed GPS, all readings look good


fazerlicourice7 (approved):
Awesome!! We're finally done!

PR #66, opened 2021-04-08
Serial debug build flag

fazerlicourice7:
just slightly modify the `SERIAL_INPUT` flag to make it more obvious what each state is.


nlautrette (approved):
Looks good

PR #67, opened 2021-04-08
Vamshi/ethernet comms

fazerlicourice7:


src/E1_coldflow/E1_coldflow.cpp
```diff
@@ -12,7 +12,7 @@
 #include <tempController.h>
 #include <batteryMonitor.h>
 
-#define SERIAL_INPUT 0 // 0 is flight config, 1 is for debug
+#define SERIAL_INPUT 1 // 0 is flight config, 1 is for debug
```
taitmate38: we should always make sure the committed version has SERIAL_INPUT set to 0, so that we can pull and upload to the flight computer without changing anything.

src/E1_coldflow/E1_coldflow.cpp
```diff
@@ -169,7 +211,7 @@ void loop() {
       Automation::detectPeaks(loxInjector, propInjector);
     }
   }
-  delay(10);
+  delay(100);
```
taitmate38: I assume this larger delay is just for testing.? & should be changed back before merging

taitmate38 (dismissed):

lib/common_fw/common_fw.cpp
```diff
@@ -7,31 +7,20 @@
...
-        }
-        debug("flushing file");
-        file.flush();
-        return true;
-    }
-    return true;
-}
+#ifdef ETH
+EthernetUDP Udp;
+unsigned int port = 6969; // try to find something that can be the same on gs
```
salsquared, line 21: nice B)

src/LAD4_1/lad4_1.cpp
```diff
@@ -117,10 +117,10 @@ void loop() {
     sensor = &sensors[j];
     sensorReadFunc(sensor->id);
     packet = make_packet(sensor->id, false);
-    Serial.println(packet);
+    Serial.print(packet);
 
     #if SERIAL_INPUT != 1
-        RFSerial.println(packet);
+        RFSerial.print(packet);
```
salsquared: Why are some of these print statements being changed from println() to print() and some arent?

src/E1_coldflow/E1_coldflow.cpp
fazerlicourice7: good catch

src/LAD4_1/lad4_1.cpp
fazerlicourice7: good catch. that's an accident

fazerlicourice7:
[Ethernet Communications](https://trello.com/c/9dje3fGk/226-ethernet-communications)

PR #68, opened 2021-04-11
Sal/rtc code

salsquared:
Added realTimeClock.h so that we can setup the RTC on the teensy and use some of the <TimeLib.h> functionality to name our files


lib/realTimeClock/realTimeClock.h
```diff
@@ -0,0 +1,39 @@
...
+
+    void init() {
+      if(timeStatus() != timeSet) 
+        setSyncProvider(RTC.get);
+    }
+
+    String getFileTime() {
+      String result = String(month()).concat("-");
+      result.concat(day());
+      result.concat("-");
```
fazerlicourice7: let's use `std::string` instead of Arduino's `String` we're trying to move in that direction.

fazerlicourice7 (changes requested):
You'll need to actually set the file name in `config.h` for each different env.
And you have merge conflicts you need to fix.

salsquared:
I also think that it is already writing to a new file at every boot

platformio.ini
```diff
@@ -22,10 +24,6 @@ lib_deps =
 	greiman/SdFat@^2.0.4
 	adafruit/Adafruit BMP3XX Library@^2.0.1
 
-[env:E1_coldflow]
-src_filter = -<*> +<E1_coldflow/>
-build_flags = -DHAL=teensy -DDEBUG ; -DSERIAL_INPUT_DEBUG
-
 [env:E1_coldflow_spoof]
```
fazerlicourice7: why'd you delete this?

src/E1_coldflow/E1_coldflow.cpp
```diff
@@ -87,6 +88,8 @@ void setup() {
   _cryoTherms.init(numCryoTherms, _cryo_boards, cryoThermAddrs, cryoTypes);
 
   Automation::init();
+  
+  RealTimeClock::init();
```
fazerlicourice7: move this above config::setup()

lib/common_fw/common_fw.h
```diff
@@ -88,6 +89,16 @@ struct sensorInfo {
...
+// namespace defining custom to_string function
+namespace patch {
+  template <typename T> 
+  std::string to_string(const T& n) {
+    std::ostringstream s;
+    s << n;
+    return s.str();
+  }
+}
+
```
fazerlicourice7: just use itoa. this creates unnecessary bloat.

fazerlicourice7 (changes requested):

platformio.ini
salsquared: was a mistake

lib/common_fw/common_fw.h
```diff
@@ -9,6 +9,7 @@
 
 #include <string>
 #include <cstring>
+#include <sstream>
```
fazerlicourice7, line 12: remove

platformio.ini
```diff
@@ -8,12 +8,14 @@
 ; Please visit documentation for the other options and examples
 ; https://docs.platformio.org/page/projectconf.html
 
+[platformio]
+
```
fazerlicourice7, line 12: remove

fazerlicourice7 (commented):
cool Looks good.
two tiny things
have you tested this out, make sure it works as expected?

fazerlicourice7:
update?

PR #70, opened 2021-04-15
Hotfire automation

daangbruce:
Hotfire automation + abort sequence

### DIFF
--- a/lib/Automation/Automation.cpp
+++ b/lib/Automation/Automation.cpp
@@ -22,19 +22,6 @@ namespace Automation {
 
  struct autoEventList* _eventList;
 
-  /* Delays during startup sequence:
-    1 - Between open pressure and open LOX Main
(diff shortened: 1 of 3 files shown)


lib/Automation/Automation.cpp
```diff
@@ -22,19 +22,6 @@ namespace Automation {
...
-    3 - Between open Prop Main and close Main Valve Arm
-  */
-  int _startupDelays[3] = {1000, 0 ,1000};
-
-  /* Delays during shutdown sequence:
-    1 - Between arm 2-way, close high pressure, close Prop and close LOX
-    2 - Between close LOX and closing Arming valve & opening Gems vent
-  */
-  int _shutdownDelays[2] = {0, 750};
-
```
fazerlicourice7, line 37: Good catch that these aren't being used

fazerlicourice7 (approved):

PR #74, opened 2021-04-23
DAQ verified + small changes

nlautrette:
- verified that ADC outputs accurate values
- verified load cell readings
- uncommented load cell stuff
- changed cryo therm types to all be K
- changed the way analog thermocouples are read to support multiple thermocouples

### DIFF
--- a/lib/ducer/ducer.h
+++ b/lib/ducer/ducer.h
@@ -87,6 +87,21 @@ namespace Ducers {
   }
 
 
+  float interpolate(long rawValue, int ptType) { // 300 psi sensor
+    double values[2][2] = { // [x, y] pairs
+      {6553,  0},
+      {58982, (double) ptType}
+    };
+    // return std::lerp(-123.89876445934394, 1131.40825, (double) rawValue / 64901);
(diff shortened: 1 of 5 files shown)


lib/ducer/ducer.h
```diff
@@ -87,6 +87,21 @@ namespace Ducers {
   }
 
 
+  float interpolate(long rawValue, int ptType) { // 300 psi sensor
+    double values[2][2] = { // [x, y] pairs
+      {6553,  0},
+      {58982, (double) ptType}
```
fazerlicourice7, line 93: ahh GG. nicely done. 
Are we sure they all line up w/ 0.5 V = 0 psi, and 4.5V = max? 
@andycate  @taitmate38  how are we doing the new high pressure sensor? does it max at 5 or 4.5?

fazerlicourice7 (commented):
Can you fix the build failures.
And I like the generalization on the ducer interpolation. But Idk if all the sensors have the same voltage ranges. I think you can do that for all of them except the high pressure sensor.

PR #75, opened 2021-04-23
Flight Computer Thermocouple Amplifier fix

taitmate38:
Found out that the issues we were having with the MCP9600 thermocouple amplifiers were due to the chips themselves - we got an older revision of the chip that had some known issues. The fix listed in the Datasheet was to slow down the speed of the I2C bus to below 85KHz. Turns out the existing method for changing speed only allows 100KHz or 400KHz, so we copied the code in that method and modified it to manually set bus speed for 50KHz for thermocouple amps. Wanted to do it as a part of library, but code was crashing when done that way (even though it was working on the DAQ - a result of some lingering pointer issue). 

But this code was tested & verified during propane fill testing.


rling101:
woohoo!

src/E1_hotfire_mvp/E1_hotfire_mvp.cpp
```diff
@@ -107,6 +107,26 @@ void setup() {
...
+
+
+  /* START SET I2C CLOCK FREQ */
+  // Wire.setClock(85000);
+  #define CLOCK_STRETCH_TIMEOUT 15000
+  IMXRT_LPI2C_t *port = &IMXRT_LPI2C1;
+  port->MCCR0 = LPI2C_MCCR0_CLKHI(55) | LPI2C_MCCR0_CLKLO(59) |
+    LPI2C_MCCR0_DATAVD(25) | LPI2C_MCCR0_SETHOLD(40);
+  port->MCFGR1 = LPI2C_MCFGR1_PRESCALE(2);
+  port->MCFGR2 = LPI2C_MCFGR2_FILTSDA(5) | LPI2C_MCFGR2_FILTSCL(5) |
```
fazerlicourice7, line 121: we need this in `config.h` and `__.cpp` ?

fazerlicourice7 (approved):

PR #79, opened 2021-04-27
added linear actuator states to AC

nlautrette:
added packet 6 to linearActuators.cpp, which reports the state of the linear actuators

### DIFF
--- a/lib/linearActuators/linearActuators.cpp
+++ b/lib/linearActuators/linearActuators.cpp
@@ -33,6 +33,9 @@ namespace LinearActuators {
 
   LinActCommand * _linActCommands[7] = {&zero, &one, &two, &three, &four, &five, &six};
 
+  uint8_t _allStates[7];
+  uint8_t _prevStates[7];
+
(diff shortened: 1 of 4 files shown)


lib/linearActuators/linearActuators.cpp
```diff
@@ -83,11 +93,33 @@ namespace LinearActuators {
...
-      data[i] = 2 * _in1PinState[i] + _in2PinState[i];
+      int state = 2 * _in1PinState[i] + _in2PinState[i];
+      data[i] = state;
+      if (state == 0 && _prevStates[i] != 0) {
+        //if previously moving forward, new state is open (1)
+        if (_prevStates[i] == 1) {
+          _allStates[i] = 1;
+        //if previously moving backward, new state is closed (0)
+        } else if (_prevStates[i] == 2) {
+          _allStates[i] = 0;
```
fazerlicourice7, line 108: What if it didn't open/close all the way.
Can we easily check if it stopped because of timer or current detection, and set appropriately?

nlautrette: if it didn't close or open all the way then the state shouldn't change, its still "somewhere in between"

PR #81, opened 2021-04-29
Changing Read Rate of Sensors

das2002:
change is made on line 163 of E1_hotfire_mvp.cpp, just added a for loop that slows down the read rate of sensors of ids 5-10 by 5 cycles (arbitrarily chosen).


fazerlicourice7:
Add .idea/ to the .gitignore file

fazerlicourice7:
And this should be a Pr into hot fire-mvp 
Not master

PR #82, opened 2021-04-30
sensor read rate change

das2002:
fixed the errors that I made in creating the pull request (added .idea/ to the gitignore and am requesting for the current branch)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,4 @@
 .pio/
 .gcc-flags.json
 .clang_complete
+.idea/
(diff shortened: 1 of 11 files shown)


fazerlicourice7:
The .idea directory is still here. you'll probably need to force git rm
Also, fix merge conflicts.

fazerlicourice7 (changes requested):
looks like the .idea dir is still there. And there are merge conflicts that need to be fixe.d

PR #83, opened 2021-07-07
Vamshi/docs

fazerlicourice7:
Started some docs, renamed some vars to be closer to what they actually do, based on discussions w/ people.


taitmate38:
out of date

PR #86, opened 2021-09-14
Hotfire3/fc amb func automation basic

taitmate38:
This branch appears to be stable enough, given that we've been using it for all water flows & cold flows that we've been doing with the cart Includes:
-  updated cart mappings
- change to how autosequences are executed to have state transitions, which control when the autosequence ends, defined as a part of the auto sequence itself, instead of elsewhere in the code.
- define flags to easily enable or disable IgniterCurrentCheck and EndFlowDetection. 
- Update solenoids library to turn on GEMS valves at the same time as the fuel and ox main valves, allowing the LEDs to give visual feedback of valve command timing as well as for camera synchronization.

### DIFF
--- a/lib/Automation/Automation.cpp
+++ b/lib/Automation/Automation.cpp
@@ -5,11 +5,16 @@
 
 #include "Automation.h"
 
+
+
 using namespace std;
 
 namespace Automation {
 
-//-----------------------Variables-----------------------
+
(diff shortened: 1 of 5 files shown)


andycate (approved):
Seems good to me

nlautrette (approved):

PR #89, opened 2021-09-18
Hotfire3/oc protection

andycate:


moose73:
This PR doesn't have the SD code, so doesn't work (it compiles fine, but packets dont get sent). We can delete this PR in a bit, just keeping it up so we have a record of changes ig

taitmate38:
Changes included in aggregate PR instead: https://github.com/Space-Enterprise-at-Berkeley/Firmware/pull/92

PR #90, opened 2021-09-19
Hotfire3/unique mac

xy3710:
Replaced hardcoded MAC with a few lines that query the Teensy's MAC registers. 

-Had to remove a bunch of SD/file stuff to get the code to build, but the meat of the unique MAC is only a few short lines.


taitmate38:
Changes included in aggregate PR instead: https://github.com/Space-Enterprise-at-Berkeley/Firmware/pull/92

PR #92, opened 2021-09-27
Firmware features verified on 9/25

taitmate38:
A bit of a big PR, but that's because it contains code from multiple features that were all tested this weekend. The features included in this PR are:

- Overcurrent protection on valves & heaters. Using limits of 2A for valves and 4A for heaters. Main goal is to prevent damage via any shorts on any of the valve channels. ([AVI-59](https://jira.berkeleyse.org/browse/AVI-59))
- Use teensy built in MAC address instead of hard-coded MAC address ([AVI-89](https://jira.berkeleyse.org/browse/AVI-89))
- Store commit hash in firmware and new packet (id 99) to allow dashboard to request commit hash & other metadata, like configured burn time ([AVI-37](https://jira.berkeleyse.org/browse/AVI-37))
- Ability to disable the collection of thermocouple packets using a new packet (id 65), to allow higher read rate of PTs and other sensors during burn ([AVI-75](https://jira.berkeleyse.org/browse/AVI-75))

Another change was to the main begin flow automation sequence. The autosequence was updated to remove any autoEvents that tried to open/close the High Pressure Solenoid or the GEMS valves (since neither are used). Both LEDs were also updated to turn on when the igniter is turned on, not when either main valve is turned on. 

This code was used during [TEST-25](https://jira.berkeleyse.org/browse/TEST-25), [TEST-26](https://jira.berkeleyse.org/browse/TEST-26), and [TEST-27](https://jira.berkeleyse.org/browse/TEST-27), and no firmware issues were noticed during any of these three cold flows.

### DIFF
--- a/lib/solenoids/solenoids.cpp
+++ b/lib/solenoids/solenoids.cpp
@@ -359,6 +359,8 @@ namespace Solenoids {
 
   int armPropane() {
     digitalWrite(lox_gems_pin, 1);
+    digitalWrite(prop_gems_pin, 1);
+    
     if (prop2_state == 0) {
       toggleProp2Way();
     } else {
@@ -369,6 +371,7 @@ namespace Solenoids {
 
   int disarmPropane() {
     digitalWrite(lox_gems_pin, 0);
+    digitalWrite(prop_gems_pin, 0);
     if (prop2_state == 1) {
       toggleProp2Way();
     } else {
@@ -411,7 +414,7 @@ namespace Solenoids {
 
   int openPropane() {
 
-    digitalWrite(prop_gems_pin, 1);
+    //digitalWrite(prop_gems_pin, 1);
 
     if (prop5_state == 0) {
       toggleProp5Way();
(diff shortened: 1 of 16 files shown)


lib/solenoids/solenoids.cpp
```diff
@@ -322,6 +358,9 @@ namespace Solenoids {
   }
 
   int armPropane() {
+    digitalWrite(lox_gems_pin, 1);
```
nlautrette, line 361: we should change these names

nlautrette (approved):

andycate (approved):
LGTM

lib/solenoids/solenoids.cpp
taitmate38: that is quite a rabbit hole to go down, given all of our "propane" names scattered throughout lol. Maybe on the next merge though

taitmate38:
okay I'll merge, in the future I'll try to make PRs of a branch like this into main more for record keeping than anything else. Let's try to make PRs of individual features into an aggregate branch like this where we leave comments and feedback,

PR #95, opened 2021-10-06
Firmware Features verified on 9/30 & 10/02 - aka code for Hotfire!

taitmate38:
This PR contains a few updates to features from the last merge that were tested during the last few cold flows done with the system. With this, we have all the functionality that we could reasonably want for hotfire, and thus this will be the last code update until then. Updates include:

New Feature:
- Finally got the event-based message packets to compile and work alongside with the rest of the changes, meaning that now when an autoEvent is defined, you can change a single boolean to true and add an integer message ID that will be sent as the data for a packet with id 57 right after that message has been executed. Found and fixed bug where id being sent was for next event instead of current event. ([AVI-79](https://jira.berkeleyse.org/browse/AVI-79))
- Updated Actuator Controller firmware to use correct configuration for INA219 current monitors, allowing reading of currents >1.21A. Also integrated use of built-in hardware MAC address. ([AVI-100](https://jira.berkeleyse.org/browse/AVI-100))
- Re-enabled and verified that the IgniterCurrentCheck will abort if the igniter see no current. Updated autoSequence to turn off flag, allowing the check to be ran multiple times without needing to power cycle, and added a message packet that is sent when an igniter Abort occurs. 
- Added teensy hardware MAC address code to DAQ 1 as well. 
- Most recent code for AC and DAQ was merged into this branch as well, meaning this branch has the most up to date code for _all boards_!

Updates:
- Updated overcurrent protection to also cover GEMS channels, since they are seeing power for LEDs. Tested and confirmed that channels get shut off and that message packets get sent. ([AVI-60](https://jira.berkeleyse.org/browse/AVI-60))
- Ability to re-enable thermocouple packets, meaning now packet 65 can be sent with a 1 or 0 to enable/disable tank thermocouples at will. ([AVI-75](https://jira.berkeleyse.org/browse/AVI-75))

Most of the changes were incorporated and used during [TEST-28](https://jira.berkeleyse.org/browse/TEST-28) on 9/30, and then all the changes were used during [TEST-30](https://jira.berkeleyse.org/browse/TEST-30) and [TEST-32](https://jira.berkeleyse.org/browse/TEST-32) on 10/02, and no firmware issues were noticed during any of these three cold flows. On 10/02 in particular, boards were powered on for >6 hours, with no concerns like crashing observed at all during that time period.


andycate (approved):
LGTM

nlautrette (approved):

PR #97, opened 2021-11-06
 Autosequence Updates for Separate Valve Actuation 

taitmate38:
Mainly updated auto sequence to have LOX open separately than Fuel, and eventPackets updated to reflect that.


taitmate38:
the other PR was merging into hotfire3/main. Made new `hotfire4/main` branch, and set it as the default branch for the repo.

PR #98, opened 2021-11-15
Hotfire4/firmware dev 11 10

taitmate38:
This PR includes changes that were made and used during acceptance testing campaign. Changes include:

- Adding detection of breakwire continuity going from closed -> open, and a flag that is set so that automation sequence can abort if there is no detected change in continuity. Enable/Disable flag is in Automation.cpp
- Updated INA219 library to correctly read bus voltages >15V - allows nominal continuity detection on 24V channels
- Added flag in Solenoids to enable or disable an LED channel being used as a breakwire - if true, the flight computer will not attempt to power the channel when the igniter is commanded on. 
- Updated automation sequence to have delay adequate for propane, as seen from testing. Last change was 165ms LOX lead

### DIFF
--- a/lib/Automation/Automation.cpp
+++ b/lib/Automation/Automation.cpp
@@ -13,6 +13,7 @@ namespace Automation {
 
 
   const bool IGNITER_DETECT_ENABLED = true;
+  const bool BREAK_DETECT_ENABLED = true;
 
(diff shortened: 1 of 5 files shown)


andycate (approved):
Looks good to me. Is this the final valve timing for hotfire?

PR #100, opened 2021-12-25
initial comms implementation

minhn02:


src/Comms.cpp
```diff
@@ -1,15 +1,31 @@
...
     uint32_t packetWaiting() {
-        return 0; // return 1 if packets waiting
+        return Udp.parsePacket();
     }
 
     void processPackets() {
-        // process all waiting packets
+        Udp.read(packetBuffer, sizeof(Packet));
+
+        Packet packet = *(struct Packet *)&packetBuffer;
```
andycate: We want to verify the checksum before we pass this part of the function. The checksum we are using is Fletcher16, and we have a function that generates it in the old code.

src/Comms.cpp
```diff
@@ -22,9 +38,15 @@ namespace Comms {
     }
 
     void emitPacket(Packet *packet) {
-        Serial.print(packet->data[0]);
-        Serial.print(packet->data[1]);
-        Serial.print(packet->data[2]);
-        Serial.println(packet->data[3]);
+        //Send over serial
+        for (int i = 0; i < packet->len; i++) {
+            Serial.println(packet->data[i]);
```
andycate: Use Serial.write instead, since it writes each byte directly to the UART port without any other processing

src/Comms.cpp
```diff
@@ -22,9 +38,15 @@ namespace Comms {
...
+        }
+        Serial.println("");
```
andycate: Use Serial.write('\n')

src/Comms.cpp
```diff
@@ -22,9 +38,15 @@ namespace Comms {
...
+
+        //Send over ethernet
+        Udp.beginPacket(Udp.remoteIP(), Udp.remotePort());
+        Udp.write(packet->data, packet->len);
```
andycate, line 109: Need to write packet id, length, and checksum first before writing the data

src/Comms.h
```diff
@@ -1,6 +1,15 @@
...
 #include <Arduino.h>
+#include <Ethernet.h>
+#include <EthernetUdp.h>
+#include <map>
+#include <string>
+
+//TODO replace with actual values
+byte mac[] = {0x00, 0x00, 0x00, 0x00, 0x00, 0x00};
+int port = 0;
+IPAddress ip(192, 168, 1, 177);
```
andycate: The old code had a function to retrieve the Mac address from the teensy. You should find that and use it for this array.

The port should be 42069 and the IP address should be 10.0.0.42

andycate (changes requested):
Good start, but there are a few changes that need to be made

src/Comms.cpp
```diff
@@ -22,9 +41,46 @@ namespace Comms {
...
+        uint8_t *ptr = (uint8_t *)&checksum;
+        packet->checksum[0] = ptr[0];
+        packet->checksum[1] = ptr[1];
+
+        //Send over serial
+        Serial.write(packet->id);
+        Serial.write(packet->len);
+        Serial.write(packet->checksum, 2);
+        Serial.write(packet->data, packet->len);
+        Serial.write("\n");
```
andycate: Change "\n" to '\n'

src/Comms.h
```diff
@@ -1,19 +1,44 @@
...
 
 #include <Arduino.h>
+#include <Ethernet.h>
+#include <EthernetUdp.h>
+#include <map>
+#include <string>
+
+byte mac[] = {0xDE, 0xAD, 0xBE, 0xEF, 0xDA, 0x02};
+int port = 42069;
+IPAddress ip(10, 0, 0, 42);
```
andycate: You should move these inside the namespace, and make them const

andycate: Also, every teensy has a unique Mac address built into the hardware. There's a function you can use to get the Mac address of the teensy, and you should use that to set this array. This is how it was done in the old code: 
image

src/Comms.h
```diff
@@ -1,19 +1,44 @@
 #pragma once
 
 #include <Arduino.h>
+#include <Ethernet.h>
+#include <EthernetUdp.h>
+#include <map>
+#include <string>
```
andycate: What's this used for?

src/Comms.cpp
```diff
@@ -22,9 +41,46 @@ namespace Comms {
     }
 
     void emitPacket(Packet *packet) {
-        Serial.print(packet->data[0]);
-        Serial.print(packet->data[1]);
-        Serial.print(packet->data[2]);
-        Serial.println(packet->data[3]);
+        //generate checksum from data and add it to the packet
+        uint16_t checksum = fletcher16(packet->data, packet->len);
```
andycate: The checksum should include the id and the len variables in the packet. Might wanna change the fletcher16 function to take in a packet pointer instead of just a byte array

src/Comms.cpp
```diff
@@ -1,15 +1,34 @@
...
 
     void processPackets() {
-        // process all waiting packets
+        Udp.read(packetBuffer, sizeof(Packet));
+
+        Packet packet = *(struct Packet *)&packetBuffer;
+
+        //call callback function
+        uint16_t checksum = *(uint16_t *)&packet.checksum;
+        if (checksum == fletcher16(packet.data, packet.len)) {
```
andycate: See my comment below too, but the checksum should include the id and the len of the packet. Might wanna just change the fletcher16 function to take in a packet pointer instead of a byte array.

andycate (changes requested):
New changes look good, just need to change slightly how the checksum is calculated

src/Comms.h
minhn02: Nothing :)

src/Comms.cpp
```diff
@@ -22,9 +41,59 @@ namespace Comms {
...
+
+        //Send over serial
+        Serial.write(packet->id);
+        Serial.write(packet->len);
+        Serial.write(packet->checksum, 2);
+        Serial.write(packet->data, packet->len);
+        Serial.write('\n');
+
+        //Send over ethernet
+        Udp.beginPacket(Udp.remoteIP(), Udp.remotePort());
```
andycate: Here, the port can just be 42069 like it is above. The IP should actually be `10.0.0.69` and `10.0.0.70`. So this code needs to actually run twice to send to both IP addresses.

src/Comms.cpp
```diff
@@ -22,9 +41,59 @@ namespace Comms {
...
+     */
+    uint16_t computePacketChecksum(Packet *packet) {
+        uint8_t vals[packet->len + 2];
+        vals[0] = packet->id;
+        vals[1] = packet-> len;
+        for (int i = 0; i < packet->len; i++) {
+            vals[i + 2] = packet->data[i];
+        }
+        return fletcher16(vals, packet->len + 2);
+    }
```
andycate: This seems like a somewhat inefficient way to do the checksum. I'd actually just modify the fletcher16 function directly. If you just set `sum1` equal to `packet->id` and `sum2` equal to `(packet-> id + packet-> len) % 256`, its the equivalent of this function.

src/Comms.cpp
```diff
@@ -22,9 +41,59 @@ namespace Comms {
...
+     */
+    uint16_t fletcher16(uint8_t *data, int len) {
+
+        uint16_t sum1 = 0;
+        uint16_t sum2 = 0;
+
+        for (int index=0; index<len; index++) {
+            if (data[index] > 0) {
+            sum1 = (sum1 + data[index]) % 255;
+            sum2 = (sum2 + sum1) % 255;
```
andycate: these should both have `% 256` instead of `% 255`

andycate (changes requested):
Just a few more nit-picky things

andycate (approved):
Looks good!

PR #106, opened 2022-02-02
V2/relay fix

taitmate38:
- Added additional valve to Valves namespace to make use of a 24V channel for igniterEnableRelay - both actuation and sampling
- Added additional step to automation to turn off igniter and new igniterEnable 0.5s after turning on, instead of having it on during full 2 second period between e-match ignition and lox main valve opening


src/FC/Automation.cpp
```diff
@@ -81,22 +81,30 @@ namespace Automation {
         DEBUG(step);
         DEBUG("\n");
         switch(step) {
-            case 0: // step 0 (actuate igniter)
+            case 0: // step 0 (enable and actuate igniter)
+                //TODO: don't even turn on igniter if igniterEnabled = False
                 if(Valves::breakWire.voltage > breakWireThreshold || !breakwireEnabled) {
+                    Valves::enableIgniter();
                     Valves::activateIgniter();
                     sendFlowStatus(0);
```
andycate: Need to edit all the sendFlowStatus function calls to use the preprocessor flags

andycate (changes requested):
Looks good, just need to use the compiler flags in the sendFlowStatus functions.

src/FC/Automation.cpp
taitmate38: wanted to start with just the one I added, but yea will change out all the others too

taitmate38:
Updated to use compiler directives instead of hard-coded constants for abort flags

PR #109, opened 2022-03-01
GPS module

athenaleong:
Added GPS module. Test code is in /tests/gps-test.

### DIFF
--- a/platformio.ini
+++ b/platformio.ini
@@ -16,51 +16,109 @@ framework = arduino
 [env:flight]
 build_flags = -DIP_ADDRESS_END=42
 src_filter = -<*> +<Common.h> +<Comms.*> +<FC/*>
+lib_deps = 
+	stevemarple/MicroNMEA@^2.0.5
+	sparkfun/SparkFun u-blox Arduino Library@^1.8.11
+	sparkfun/SparkFun u-blox GNSS Arduino Library@^2.2.0
 
 [env:flight-debug]
 build_flags = -DIP_ADDRESS_END=42 -DDEBUG_MODE
 src_filter = -<*> +<Common.h> +<Comms.*> +<FC/*>
+lib_deps = 
+	stevemarple/MicroNMEA@^2.0.5
+	sparkfun/SparkFun u-blox Arduino Library@^1.8.11
+	sparkfun/SparkFun u-blox GNSS Arduino Library@^2.2.0
 
 [env:ac1]
 build_flags = -DIP_ADDRESS_END=21
 src_filter = -<*> +<Common.h> +<Comms.*> +<AC/*>
+lib_deps = 
+	stevemarple/MicroNMEA@^2.0.5
+	sparkfun/SparkFun u-blox Arduino Library@^1.8.11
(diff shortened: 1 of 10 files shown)


platformio.ini
```diff
@@ -16,51 +16,109 @@ framework = arduino
 [env:flight]
 build_flags = -DIP_ADDRESS_END=42
 src_filter = -<*> +<Common.h> +<Comms.*> +<FC/*>
+lib_deps = 
+	stevemarple/MicroNMEA@^2.0.5
+	sparkfun/SparkFun u-blox Arduino Library@^1.8.11
+	sparkfun/SparkFun u-blox GNSS Arduino Library@^2.2.0
```
andycate, line 22: For these dependencies, can you download them and put them into the `lib` folder? That way we don't need to have external dependencies

src/FC/GPS.cpp
```diff
@@ -0,0 +1,84 @@
+#include "GPS.h"
+#include <Wire.h>
+
+
+namespace GPS {
+
+    SFE_UBLOX_GNSS myGNSS;
+    uint32_t gpsUpdatePeriod = 40 * 1000;
+
+    Comms::Packet latLongPacket = {.id = 11};
```
andycate: The ID of the gps packet is `6`, according to the [E-1 Design spreadsheet](https://docs.google.com/spreadsheets/d/1xNmjrb5a28FfVg-MG9jUBRiPyfdkjo7HiqPWP41N_IQ/edit#gid=1524498060)

src/FC/GPS.cpp
```diff
@@ -0,0 +1,84 @@
...
+    double latitude;
+    double longitude;
+
+
+    Comms::Packet auxPacket = {.id = 12};
```
andycate: Packet 12 is for something else rn, can you add a line to the E-1 Design spreadsheet for a new packet? Maybe ID 7, since thats unused right now

src/FC/GPS.cpp
```diff
@@ -0,0 +1,84 @@
...
+        Wire.begin();
+        Serial.println("GPS setup");
+        if (!myGNSS.begin()) {
+            Serial.println("GPS failed to start");
+        } else {
+            Serial.println("GPS started");
+        }
+        // Version 2 - UBX
+        myGNSS.setI2COutput(COM_TYPE_UBX);   //Set the I2C port to output UBX only (turn off NMEA noise)
+        myGNSS.saveConfigSelective(VAL_CFG_SUBSEC_IOPORT); //Save (only) the communications port settings to flash and BBR
```
andycate: On the flight computer, the GPS is connected to the teensy using SPI, so can you set up the SPI interface here instead of I2C? Also, don't use Serial.print for logging. If you want to log messages for debugging purposes, use the `DEBUG("<message>");` macro

andycate (changes requested):
Looking good, but need to adapt the code to communicate with the GPS over SPI. Should be relatively easy to change, given that the library has SPI bus support built in.

taitmate38:
We should have a separate v2/lad5 branch that we are PR-ing all these branches into

